PR #3, opened 2021-05-06
Opaque slice op

Speierers:
This PR is about:

- [x] improving the `ek::opaque` routine to make it compatible with static array and struct input values.
- [x] evaluate literal array with `array = ek::opaque(array)`
- [x] re-implement the `ek::slice` routine to access slices of various data structures (similar to what we had in the previous version of enoki). The default `index=-1` argument will check that the given data structure as a width of 1.
- [x] implement basic unit tests

### DIFF
--- a/include/enoki/array_router.h
+++ b/include/enoki/array_router.h
@@ -1190,27 +1190,37 @@ decltype(auto) migrate(const T &value, TargetType target) {
     }
 }
 
-template <typename T>
-decltype(auto) slice(const T &value, size_t index = -1) {
-    if (index == (size_t) -1) {
-        if(width(value) > 1)
-            enoki_raise("slice(): variable contains more than a single entry!");
-        index = 0;
-    }
+template <typename ResultType = void, typename T>
+decltype(auto) get_slice(const T &value, size_t index = -1) {
     if constexpr (array_depth_v<T> > 1) {
-        using Value = decltype(slice(value.entry(0), index));
+        using Value = std::decay_t<decltype(get_slice(value.entry(0), index))>;
         using Result = typename T::template ReplaceValue<Value>;
         Result result;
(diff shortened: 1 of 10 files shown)


src/python/bind.h
```diff
@@ -159,7 +159,7 @@ auto bind_full(py::class_<Array> &cls, bool scalar_mode = false) {
     cls.attr("full_") = py::cpp_function(
         [](Scalar v, size_t size) { return ek::full<Array>(v, size); });
     cls.attr("opaque_") = py::cpp_function(
-        [](Scalar v, size_t size) { return ek::opaque<Array>(v, size); });
+        [](Array v, int size) { return ek::opaque(v, (size_t) size); });
```
wjakob, line 162: Not sure about the point of the change to `int`, if you in any case cast to `size_t` just below. If you need the argument to be signed, why not `ssize_t`?

include/enoki/array_router.h
```diff
@@ -1180,6 +1190,31 @@ decltype(auto) migrate(const T &value, TargetType target) {
     }
 }
 
+template <typename T>
+decltype(auto) slice(const T &value, size_t index = -1) {
+    if (index == (size_t) -1) {
+        if(width(value) > 1)
```
wjakob, line 1196: space after `if`

wjakob (approved):
Looks reasonable thanks. Just two very minor comments.

src/python/bind.h
Speierers: Indeed I need the argument to be signed otherwise I can't pass `-1` (or the default). Maybe this is more a pybind11 issue, but it complains about a type error.

I will just use `ssize_t` for this then.

Speierers:
Merged already ;)

PR #11, opened 2021-09-30
Tensor: fix `select` return type

merlinND:
The result of all select operations on Tensor types was accidentally being cast to integer.
Also adds a unit test covering this bug.


wjakob:
oops, what a weird issue with the code. Not sure what I was thinking.

PR #17, opened 2021-11-25
Tensor: fix accidental migrations due to `hasattr()`

merlinND:
The Tensor constructor from a Python object checks whether that object makes available an `__array_interface__` with `hasattr()`.

But `hasattr()` actually *executes* the `__array_interface__` property in order to determine whether it is available (https://stackoverflow.com/a/30143990/3792942)

If the Python object was an Enoki array, in many cases, `__array_interface__` actually needs to migrate the array to the host to make its data available.
This lead to many accidental migrations to host memory when constructing tensors from other Enoki arrays.

Ideally, we could find a way to allow `hasattr` while avoiding actual migrations, but in the meantime this fix simply avoids the `hasattr` call for Enoki arrays.


wjakob:
Ouch -- thanks Merlin!

PR #18, opened 2021-12-03
Texture: fix AD support

merlinND:
- Enoki-side storage is now a tensor instead of an array
- `.value()` and `.tensor()` both return references so that `enable_grad()`, etc calls made on them have the expected effect
- Fixed logic checking for invalid `m_value` to trigger migration
- Updated unit test


include/enoki/texture.h
```diff
@@ -272,6 +281,8 @@ template <typename Value, size_t Dimension> class Texture {
             m_size *= m_shape[i];
         }
         m_shape[Dimension] = (uint32_t) channels;
+        m_value = TensorXf(empty<Storage>(m_size), Dimension + 1, m_shape);
```
merlinND, line 284: I think it's important to set the tensor dimensions correctly here, but I didn't know how to do it without also creating an array of the same size.

include/enoki/texture.h
```diff
@@ -119,44 +123,51 @@ template <typename Value, size_t Dimension> class Texture {
             if (m_migrate) {
                 // Fully migrate to texture memory, set m_value to zero
                 if constexpr (IsDiff)
-                    m_value = replace_grad(Storage(0), value);
+                    m_value.array() = replace_grad(Storage(0), value);
```
wjakob, line 126: Actually, can we use `Storage()` instead of `Storage(0)` to indicate that the data has been migrated? The zero initialization was my mistake, that actually seems a little misleading.

include/enoki/texture.h
```diff
@@ -52,6 +52,7 @@ template <typename Value, size_t Dimension> class Texture {
      */
     Texture(size_t shape[Dimension], size_t channels, bool migrate = true) {
         init(shape, channels, migrate);
+        m_value.array() = Storage(0);
```
wjakob, line 55: Zero initialization here makes sense in this case, because the texture is initially empty. However, I have a change request regarding zero-initialization below.

include/enoki/texture.h
```diff
@@ -119,44 +123,51 @@ template <typename Value, size_t Dimension> class Texture {
...
+    }
+
+    TensorXf &tensor() const {
         if constexpr (IsCUDA) {
             if (m_migrate) {
-                Storage result = empty<Storage>(m_size);
-                jit_cuda_tex_memcpy_t2d(Dimension, m_shape, m_shape[Dimension],
-                                        m_handle, result.data());
-                return result;
+                if (m_value.array().size() != m_size) {
```
wjakob, line 155: `.empty()`?

include/enoki/texture.h
```diff
@@ -119,44 +123,51 @@ template <typename Value, size_t Dimension> class Texture {
...
+                    Storage primal = empty<Storage>(m_size);
+                    jit_cuda_tex_memcpy_t2d(Dimension, m_shape, m_shape[Dimension],
+                                            m_handle, primal.data());
+                    if constexpr (IsDiff) {
+                        m_value.array() = replace_grad(primal, m_value.array());
+                    }
```
wjakob, line 161: remove braces in this if statement (just two 1-liners)

include/enoki/texture.h
```diff
@@ -249,14 +261,14 @@ template <typename Value, size_t Dimension> class Texture {
             Array<Value, 4> result = eval_cuda(pos, active);
 
             if constexpr (IsDiff) {
-                if (grad_enabled(m_value, pos)) {
```
wjakob, line 252: oops, does grad_enabled not work for tensors?

include/enoki/texture.h
wjakob: using `empty` is suboptimal here, because it essentially wraps `cudaMalloc()`. Can you use `ek::zero`? In this case, Enoki creates an unevaluated literal constant array that doesn't use any GPU memory. In that case, the `m_value.array() = Storage(0);` in the `Texture` constructor can also go away.

include/enoki/texture.h
```diff
@@ -272,6 +284,8 @@ template <typename Value, size_t Dimension> class Texture {
             m_size *= m_shape[i];
         }
         m_shape[Dimension] = (uint32_t) channels;
+        m_value = TensorXf(empty<Storage>(m_size), Dimension + 1, m_shape);
+        m_value.array() = Storage(0);
```
wjakob, line 288: This line can then be removed also

include/enoki/texture.h
wjakob: (we really do want enoki to error out with an "arithmetic involving unitialized value" exception if somebody then tries to access the underlying Enoki array)

wjakob (changes requested):
Hi Merlin, this looks great! I had a few small change requests, feel free to merge directly after applying them.

include/enoki/texture.h
wjakob: Actually, this line can be removed after another requested change (see below)

include/enoki/texture.h
wjakob: (the zero-initialization of .array())

include/enoki/texture.h
merlinND: Added `grad_enabled` and `set_grad_enabled` support for `Tensor` in C++, let me know if it's not done in the right place (I could also have added `set_grad_enabled_` directly in the `Tensor` class).

I've also added `zero_`, but it will only work for sizes `0` and `1` since there's no good mechanism to pass the full shape in C++ I think.

include/enoki/texture.h
merlinND: I've tried to revert to uninitialized storage, but the following issue comes up: when `migrate=true` and gradients are enabled, `eval()` then tries to gather from `m_value.array()`, which is illegal.

My understanding from your original code was that the CUDA lookup would get the actual values, and then the call to `eval_enoki()` would gather from a legal zero-valued array to get the right gradients. The Enoki lookup would get optimized away, except for the AD part.

The Enoki lookup was already tested explicitly in your original code:
https://github.com/wjakob/enoki/blob/194de8397cd8e4ee1f5866cc1f5bdf2c05c65d70/tests/texture.cpp#L128


Do you want `eval_enoki()` to trigger `m_value` to be filled, or should we keep the zero initialization instead?
Note that unless there's a bug, accessing `.value()` or `.tensor()` _will_ fill the storage, so users should never see the uninitialized / zero-valued array anyway.

wjakob (approved):
Hi Merlin,

that's actually a great point. We want `gather()` from the migrated array to succeed, so it *needs* to have some value. Not sure if I was thinking of that case when developing the code originally, but it makes sense ;-) So please ignore my previous request to initialize to `Storage()`.

Thanks,
Wenzel

wjakob:
LGTM!

wjakob:
Thanks @merlinND !

PR #22, opened 2022-01-03
Texture: fix move semantics

njroussel:
The move constructor and the move assignment operator would not even compile :sweat:
This PR fixes both methods and adds rudimentary tests cases which should at least guarantee that they compile.


wjakob (approved):
oops :D -- feel free to merge directly if the CI passes.

PR #23, opened 2022-01-03
Cubic bspline hessian matrix & resolution update

ziyi-zhang:
Simple implementation of `eval_cubic_hessian()`.
Compatible with Nicolas's changes about wrapping mode and others.

- It returns **both positional "gradient" and "hessian"**. 
They have quite some common computation. But consequently there is no interface to compute only the "hessian".
The return type is a ~`std::tuple`~`std::pair`.

- Now all the gradients/hessian return value has **the volume resolution** multiplied.
Prior to this PR, we assume the volume has unit size (because the query point is in [0, 1]^d). I need to manually multiply the shape with the returned "gradient" in my volume optimization.
But this becomes too complicated for "hessian" matrix, where each cell needs a different multiplier. So I moved this "unit size -> volume shape transformation" jacobian inside.
This change applies to the explicit gradient computation, explicit hessian computation, and implicit AD gradient computation.

### DIFF
--- a/include/enoki/texture.h
+++ b/include/enoki/texture.h
@@ -11,6 +11,7 @@
 */
 
 #include <array>
+#include <utility>
 #include <enoki-jit/texture.h>
 #include <enoki/dynamic.h>
 #include <enoki/idiv.h>
@@ -339,7 +340,7 @@ public:
     }
 
     /**
-     * \brief Helper function to evaluate a clamped cubic B-Spline interpolant
+     * \brief Helper function to evaluate a cubic B-Spline interpolant
      *
      * This is an implementation detail and should only be called by the \ref
      * eval_cubic() function to construct an AD graph. When only the cubic
@@ -356,11 +357,7 @@ public:
         using InterpIdx = uint32_array_t<InterpOffset>;
 
         PosF pos_(pos);
-        // This multiplication should not be recorded in the AD graph
         PosF pos_f = fmadd(pos_, PosF(m_shape_opaque), -.5f);
-        if constexpr (IsDiff)
-            if (grad_enabled(pos))
-                pos_f = replace_grad(pos_f, pos_);
         PosI pos_i = floor2int<PosI>(pos_f);
 
         // `offset[k]` controls the k-th offset for any dimension.
@@ -425,8 +422,7 @@ public:
     }
 
     /**
-     * \brief Evaluate a clamped cubic B-Spline interpolant represented by this
-     * texture
+     * \brief Evaluate a cubic B-Spline interpolant represented by this texture
      *
      * Intead of interpolating the texture via B-Spline basis functions, the
      * implementation transforms this calculation into an equivalent weighted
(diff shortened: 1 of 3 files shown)


include/enoki/texture.h
```diff
@@ -648,9 +651,186 @@ template <typename Value, size_t Dimension> class Texture {
         #undef EK_TEX_CUBIC_GATHER
         #undef EK_TEX_CUBIC_ACCUM
 
+        // transform volume from unit size to its resolution
+        PosF resf = PosF(m_shape_opaque);
```
wjakob, line 655: You can define this once above and reference it. For example, this variable is used to compute `pos_f` at the beginning of the function.

include/enoki/texture.h
```diff
@@ -11,6 +11,7 @@
 */
 
 #include <array>
+#include <tuple>
```
wjakob, line 14: It may be non-obvious, but `<tuple>` is one of those C++ headers that we try to avoid if at all possible (the implementation of this type is surprisingly large/complex).

What you need below is simply a `std::pair<>`, which is much lighter.

include/enoki/texture.h
```diff
@@ -356,11 +357,7 @@ template <typename Value, size_t Dimension> class Texture {
         using InterpIdx = uint32_array_t<InterpOffset>;
 
         PosF pos_(pos);
-        // This multiplication should not be recorded in the AD graph
         PosF pos_f = fmadd(pos_, PosF(m_shape_opaque), -.5f);
-        if constexpr (IsDiff)
```
wjakob, line 361: I'm curious why this change was needed?

include/enoki/texture.h
```diff
@@ -546,8 +542,15 @@ template <typename Value, size_t Dimension> class Texture {
...
 
-    /// Evaluate the positional gradient of a clamped cubic B-Spline from the
-    /// explicit differentiated basis functions
+    /**
+     * \brief Evaluate the positional gradient of a cubic B-Spline
+     *
+     * This implementation computes the result directly from explicit
+     * differentiated basis functions. It has no autodiff support.
+     * 
+     * The volume is assumed to have unit size. Otherwise, the resolution
```
wjakob, line 551: `Otherwise, the resulting gradient should be multiplied by the spatial extents.`

wjakob (commented):
Looks great @ziyi-zhang !

tests/texture.cpp
```diff
@@ -394,13 +394,52 @@ ENOKI_TEST(test11_cubic_grad_pos) {
...
     assert(ek::allclose(grad_64[0][0], ref_grad[0], 1e-5f, 1e-5f));
     assert(ek::allclose(grad_64[1][0], ref_grad[1], 1e-5f, 1e-5f));
     assert(ek::allclose(grad_64[2][0], ref_grad[2], 1e-5f, 1e-5f));
     assert(ek::allclose(grad_ad, ref_grad, 1e-5f, 1e-5f));
 }
 
-ENOKI_TEST(test12_move_assignment) {
+ENOKI_TEST(test12_cubic_hessian_pos) {
+    using TensorXf = Tensor<Float>;
+    using Uint32 = ek::uint32_array_t<Float>;
```
wjakob, line 406: We usually name this type `UInt32` (capital `I`)

include/enoki/texture.h
ziyi-zhang: This function is used to replace the AD graph of positional gradients. Now we desire the multiplication with `PosF(m_shape_opaque)` to be recorded. 
The change also applies to `eval_cubic_grad()` and `eval_cubic_hessian()`.

Why make this change (result mult with shape)?
(1) If we use the grad/hessian for volume optimization, this step is always needed. I don't know any use case of grad/hessian where this multiplication is not desired.
(2) for the gradient, one can simply do `grad * shape`. But for hessian, each cell in the matrix has a different multiplier, so it might be more convenient to move this step inside the function call.

include/enoki/texture.h
```diff
@@ -11,6 +11,7 @@
 */
 
 #include <array>
+#include <utility>
```
Speierers, line 14: What is this needed for?

Speierers (commented):
Looking great to me!

include/enoki/texture.h
ziyi-zhang: std::pair
somehow it's defined in `utility`

njroussel (commented):
Looks good to me :100: !

Speierers:
Thanks @ziyi-zhang

PR #24, opened 2022-01-07
Texture: add support for N-channel textures

njroussel:
With this PR `enoki::Texture` can now handle textures with an arbitrary number of channels.

All evaluation interfaces were changed, they now take an output parameter that is used to write the evaluation's result to. The python bindings for the evaluation methods still directly return the result, however these are now returned as python list of length N for a N-channel texture.

Tests were added to cover the new interfaces and verify the underlying implementation in `enoki-jit` when using CUDA texture objects.

:warning: This PR is already rebased on Ziyi's work in https://github.com/wjakob/enoki/pull/23 (not merged yet).

### DIFF
--- a/include/enoki/texture.h
+++ b/include/enoki/texture.h
@@ -40,12 +40,16 @@ public:
     static constexpr bool IsCUDA = is_cuda_array_v<Value>;
     static constexpr bool IsDiff = is_diff_array_v<Value>;
     static constexpr bool IsDynamic = is_dynamic_v<Value>;
+    // Only single-precision floating-point CUDA textures are supported
+    static constexpr bool HasCudaTexture =
+        std::is_same_v<scalar_t<Value>, float>;
 
     using Int32 = int32_array_t<Value>;
     using UInt32 = uint32_array_t<Value>;
-    using UInt64 = uint64_array_t<Value>;
     using Mask = mask_t<Value>;
-
+    using PosF = Array<Value, Dimension>;
+    using PosI = int32_array_t<PosF>;
+    using ArrayX = DynamicArray<Value>;
     using Storage = std::conditional_t<IsDynamic, Value, DynamicArray<Value>>;
     using TensorXf = Tensor<Storage>;
 
@@ -112,7 +116,6 @@ public:
         m_handle = other.m_handle;
         other.m_handle = nullptr;
         m_size = other.m_size;
-        m_handle_opaque = std::move(other.m_handle_opaque);
         m_shape_opaque = std::move(other.m_shape_opaque);
         m_value = std::move(other.m_value);
         for (size_t i = 0; i < Dimension; ++i)
@@ -124,14 +127,11 @@ public:
     }
 
     Texture &operator=(Texture &&other) {
-        if constexpr (IsCUDA) {
+        if constexpr (IsCUDA)
             jit_cuda_tex_destroy(m_handle);
-            m_handle = nullptr;
-        }
         m_handle = other.m_handle;
         other.m_handle = nullptr;
         m_size = other.m_size;
-        m_handle_opaque = std::move(other.m_handle_opaque);
         m_shape_opaque = std::move(other.m_shape_opaque);
         m_value = std::move(other.m_value);
         for (size_t i = 0; i < Dimension; ++i)
@@ -151,7 +151,7 @@ public:
             jit_cuda_tex_destroy(m_handle);
     }
 
-    /// Return the CUDA handle (cudaTextureObject_t). NULL on all other backends
+    /// Return the CUDA handle (EnokiCudaTexture*). NULL on all other backends
     const void *handle() const { return m_handle; }
 
     /// Return the texture dimension plus one (for the "channel dimension")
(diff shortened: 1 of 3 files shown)


include/enoki/texture.h
```diff
@@ -286,16 +281,17 @@ template <typename Value, size_t Dimension> class Texture {
             pos_i_w = wrap(pos_i_w);
             InterpIdx idx = index(pos_i_w);
 
-            Array<Value, 4> result(0);
+            for (uint32_t ch = 0; ch < channels; ++ch)
+                out[ch] = zero<Value>();
```
Speierers, line 285: `empty<>` might be enough here if we know that all entries will be overwritten?

include/enoki/texture.h
```diff
@@ -335,21 +329,25 @@ template <typename Value, size_t Dimension> class Texture {
...
+            eval_cuda(pos, out, active);
 
             if constexpr (IsDiff) {
                 if (grad_enabled(m_value, pos)) {
-                    Array<Value, 4> result_diff = eval_enoki(pos, active);
-                    result = replace_grad(result, result_diff);
+                    const size_t channels = m_value.shape(Dimension);
+
+                    DynamicArray<Value> out_enoki =
+                        empty<DynamicArray<Value>>(channels);
```
Speierers, line 342: maybe use `std::vector` or `ek::vector` instead of `DynamicArray`. It probably works but wasn't really meant to be used in this way.

include/enoki/texture.h
```diff
@@ -458,23 +450,23 @@ template <typename Value, size_t Dimension> class Texture {
...
-    Array<Value, 4> eval_cubic(const Array<Value, Dimension> &pos,
-                               Mask active = true,
-                               bool force_enoki = false) const {
+    void eval_cubic(const Array<Value, Dimension> &pos, Value *out,
+                    Mask active = true, bool force_enoki = false) const {
         using PosF = Array<Value, Dimension>;
         using PosI = int32_array_t<PosF>;
         using Array3 = Array<Value, 3>;
-        using Array4 = Array<Value, 4>;
+        using ArrayX = DynamicArray<Value>;
```
Speierers, line 458: ditto

include/enoki/texture.h
```diff
@@ -500,80 +492,99 @@ template <typename Value, size_t Dimension> class Texture {
...
+    void eval_cubic_grad(const Array<Value, Dimension> &pos,
+                         Array<Value, Dimension> *out,
+                         Mask active = true) const {
         using PosF = Array<Value, Dimension>;
         using PosI = int32_array_t<PosF>;
         using Array4 = Array<Value, 4>;
+        using ArrayX = DynamicArray<Value>;
         using InterpOffset = Array<Int32, ipow(4, Dimension)>;
         using InterpPosI = Array<InterpOffset, Dimension>;
         using InterpIdx = uint32_array_t<InterpOffset>;
```
Speierers, line 584: Some of thos aliases are used in many places. We should move them to the top of the class declaration.

Speierers (commented):
Great work @njroussel ! I did a first pass over the PR, here are a few comments.

include/enoki/texture.h
```diff
@@ -36,7 +37,8 @@ enum class WrapMode : uint32_t {
 
 template <typename Value, size_t Dimension> class Texture {
 public:
-    static constexpr bool IsCUDA = is_cuda_array_v<Value>;
+    static constexpr bool IsCUDA =
+        is_cuda_array_v<Value> && std::is_same_v<scalar_t<Value>, float>;
```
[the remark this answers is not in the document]
njroussel: What I realized  is that the CUDA textures in `enoki-jit` can only handle single-precision floating point numbers and there was nothing enforcing or at least checking for that in `Enoki::Texture`.

Should I just do these checks individually in `eval()` and `eval_cuda`?

include/enoki/texture.h
njroussel: Although they'll be overwritten, we need them to be initialized at 0, because in `EK_TEX_ACCUM()` we're accumulating them with their previous value: 

` out[ch] = fmadd(gather<Value>(m_value.array(), index_ + ch, active), weight_, out[ch]);      
`

wjakob: @Speierers , Emtpy is a wrapper around `malloc()` (at least for JIT arrays) so best to be used sparingly. Zero-initialization is cheap since it just creates a literal constant variable that doesn't use any storage (other than a hash table entry).

The JIT won't actually perform an addition if you accumulate into a variable that is currently a literal zero constant.

include/enoki/texture.h
wjakob: @Speierers I think `DynamicArray` could be good choice here. It will be mapped to the Python type `enoki.cuda.ad.ArrayXf` (we have bindings for 0/1/2/3/4D fixed size arrays, and a generic one that is indeed based on `DynamicArray`)

include/enoki/texture.h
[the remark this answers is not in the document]
Speierers: Yes I think that would be more appropriate :)

wjakob:
@njroussel, I merged the Enoki-JIT part. Does this need adaptations? (e.g. to point to the updated Enoki-JIT repo)

njroussel:
Yup, I updated the submodule. Ready to go now :rocket:

wjakob:
Excellent, thank you Nicolas!

PR #25, opened 2022-01-18
Texture: add methods to fetch texels referenced in linear interpolation

njroussel:
This PR adds `eval_fetch_enoki()`, `eval_fetch_cuda()` and `eval_fetch()`. These methods fetch the texels referenced in the linear interpolation of a given lookup position. 

`eval_fetch_cuda()` makes use of specific PTX instructions for 2D textures. For 1D and 3D textures, the method performs a normal texture lookup (`eval_cuda()`) at the center of every needed texel.

New tests are added for these methods. They cover gradient computations, GPU migration behaviors, and expected outputs in 1D/2D/3D.

***
  
  
Bug fixes & others:
- Fixes a bug in `eval_cubic` where the output could be wrong for textures with more than one channel.
- Modifies all methods where the access/traversal pattern did not correspond to the underlying storage layout of the texture data.

### DIFF
--- a/include/enoki/texture.h
+++ b/include/enoki/texture.h
@@ -353,8 +353,8 @@ public:
     }
 
     /**
-     * \brief Fetch the texels that would be reference in a CUDA texture lookup
-     * with linear interpolation without actually performing this interpolation
+     * \brief Fetch the texels that would be referenced in a CUDA texture lookup
+     * with linear interpolation without actually performing this interpolation.
      *
      * This is an implementation detail, please use \ref eval_fetch() that may
      * dispatch to this function depending on its inputs.
@@ -375,7 +375,7 @@ public:
                     fetch_pos = pos_f + PosF(ix);
                     fetch_pos *= inv_shape;
 
-                    eval_cuda(fetch_pos, out[ix]);
+                    eval_cuda(fetch_pos, out[ix], active);
                 }
             } else if constexpr (Dimension == 2) {
                 uint32_t pos_idx[Dimension];
@@ -406,7 +406,7 @@ public:
                             fetch_pos = pos_f + PosF(ix, iy, iz);
                             fetch_pos *= inv_shape;
 
-                            eval_cuda(fetch_pos, out[index]);
+                            eval_cuda(fetch_pos, out[index], active);
                         }
                     }
                 }
@@ -420,8 +420,8 @@ public:
     }
 
     /**
-     * \brief Fetch the texels that would be reference in a texture lookup with
-     * linear interpolation without actually performing this interpolation
+     * \brief Fetch the texels that would be referenced in a texture lookup with
+     * linear interpolation without actually performing this interpolation.
      *
      * If the texture data is fully migrated to the GPU, this method will return
      * zeroes.
@@ -453,8 +453,8 @@ public:
     }
 
     /**
-     * \brief Fetch the texels that would be reference in a texture lookup with
-     * linear interpolation without actually performing this interpolation
+     * \brief Fetch the texels that would be referenced in a texture lookup with
+     * linear interpolation without actually performing this interpolation.
      *
      * This function dispatches to \ref eval_fetch_enoki() or \ref
      * eval_fecth_cuda() depending on whether or not CUDA is available. If
--- a/include/enoki/texture.h
+++ b/include/enoki/texture.h
@@ -233,8 +233,8 @@ public:
 
         if constexpr (IsCUDA && HasCudaTexture) {
             uint32_t pos_idx[Dimension];
-            uint32_t *out_idx = (uint32_t *) alloca(channels * sizeof(uint32_t));
-
+            uint32_t *out_idx =
+                (uint32_t *) alloca(channels * sizeof(uint32_t));
             for (size_t i = 0; i < Dimension; ++i)
                 pos_idx[i] = pos[i].index();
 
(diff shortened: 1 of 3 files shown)


include/enoki/texture.h
```diff
@@ -304,17 +304,17 @@ template <typename Value, size_t Dimension> class Texture {
                 EK_TEX_ACCUM(idx.y(), w1.x());
             } else if constexpr (Dimension == 2) {
                 EK_TEX_ACCUM(idx.x(), w0.x() * w0.y());
-                EK_TEX_ACCUM(idx.y(), w0.x() * w1.y());
-                EK_TEX_ACCUM(idx.z(), w1.x() * w0.y());
+                EK_TEX_ACCUM(idx.y(), w1.x() * w0.y());
```
wjakob, line 307: Curious about these ordering changes, why are they needed?

include/enoki/texture.h
```diff
@@ -412,18 +551,19 @@ template <typename Value, size_t Dimension> class Texture {
...
+            Array4 wx = compute_weight(0),
+                   wy = compute_weight(1);
+            for (uint32_t iy = 0; iy < 4; iy++)
+                for (uint32_t ix = 0; ix < 4; ix++)
+                    EK_TEX_CUBIC_ACCUM(idx[iy * 4 + ix], wx[ix] * wy[iy]);
         } else if constexpr (Dimension == 3) {
             Array4 wx = compute_weight(0),
                    wy = compute_weight(1),
                    wz = compute_weight(2);
-            for (uint32_t ix = 0; ix < 4; ix++)
```
wjakob, line 423: oops, good catch!

wjakob:
This looks great. What a fancy texture class we have now 🙇

wjakob (approved):

include/enoki/texture.h
njroussel: We now have this snippet in almost all methods:
```
InterpPosI pos_i_w = interp_positions<PosI, Length>(offset, pos_i);  
pos_i_w = wrap(pos_i_w);  
InterpIdx idx = index(pos_i_w);   
```

The `interp_positions()` method generates all of the 2D integer positions of the texels that are referenced in the interpolant's computation. I realized that the order of these integer positions was "wrong": it produced integer indices (when passed to `index()`) that weren't strictly increasing and so the memory access pattern wasn't ideal. 

I changed the `interp_positions()` method and hence all the changes across all methods that directly call it.

wjakob: Got it--that makes sense, thanks!

include/enoki/texture.h
```diff
@@ -352,6 +352,145 @@ template <typename Value, size_t Dimension> class Texture {
         }
     }
 
+    /**
+     * \brief Fetch the texels that would be reference in a CUDA texture lookup
+     * with linear interpolation without actually performing this interpolation
```
Speierers, line 357:
```suggestion
     * with linear interpolation without actually performing this interpolation.
```

include/enoki/texture.h
```diff
@@ -352,6 +352,145 @@ template <typename Value, size_t Dimension> class Texture {
...
+        } else {
+            (void) pos; (void) active;
+            for (size_t i = 0; i < ipow(2, Dimension); ++i)
+                for (size_t ch = 0; ch < channels; ++ch)
+                    out[i][ch] = zero<Value>();
+        }
+    }
+
+    /**
+     * \brief Fetch the texels that would be reference in a texture lookup with
```
Speierers, line 423:
```suggestion
     * \brief Fetch the texels that would be referenced in a texture lookup with
```

include/enoki/texture.h
```diff
@@ -352,6 +352,145 @@ template <typename Value, size_t Dimension> class Texture {
...
+     * linear interpolation without actually performing this interpolation
```
Speierers, line 424:
```suggestion
     * linear interpolation without actually performing this interpolation.
```

include/enoki/texture.h
```diff
@@ -352,6 +352,145 @@ template <typename Value, size_t Dimension> class Texture {
         }
     }
 
+    /**
+     * \brief Fetch the texels that would be reference in a CUDA texture lookup
+     * with linear interpolation without actually performing this interpolation
+     *
+     * This is an implementation detail, please use \ref eval_fetch() that may
+     * dispatch to this function depending on its inputs.
```
Speierers, line 360: Isn't this method `private` anyway?

include/enoki/texture.h
```diff
@@ -352,6 +352,145 @@ template <typename Value, size_t Dimension> class Texture {
...
+        const uint32_t channels = (uint32_t) m_value.shape(Dimension);
+        for (size_t i = 0; i < InterpOffset::Size; ++i)
+            for (uint32_t ch = 0; ch < channels; ++ch)
+                out[i][ch] =
+                    gather<Value>(m_value.array(), idx[i] + ch, active);
+    }
+
+    /**
+     * \brief Fetch the texels that would be reference in a texture lookup with
+     * linear interpolation without actually performing this interpolation
```
Speierers, line 457: ditto

Speierers (commented):
Just a few minor comments on my side. Great work @njroussel !

include/enoki/texture.h
njroussel: I hadn't put too much thought into it, and set it up similarly to `eval_enoki()`, `eval_cuda()` and `eval()` which are all `public`.
The only reason I can see for keeping these public is that the `_cuda` versions have a loss of precision (GPU hardware interpolation isn't exactly 32-bit) and specific use-cases might need this fine control.

Also, it helps for testing if we don't want to setup a `friend`-system :sweat_smile:
Should I keep them `public`?

Speierers: Okay, I don't have a strong preference. 😉

PR #49, opened 2022-07-26
Generate Stubs

DoeringChristian:
I have adapted the generate_stub_files.py file from mitsuba3 to work with drjit.
I'm not experienced in CMake so I don't think my solution is that good.

### DIFF
--- a/resources/generate_stub_files.py
+++ b/resources/generate_stub_files.py
@@ -1,0 +1,331 @@
+#!/usr/bin/env python
+"""
+Usage: generate_stub_files.py {dest_dir} {drjit_path}
+
+This script generates stub files for Python type information for the `m̀itsuba`
+package.  It recursively traverses the `drjit` package and writes all the
+objects (clases, methods, functions, enums, etc.) it finds to the `dest_dir`
+folder. The stub files contain both the signatures and the docstrings of the
+objects. The second argument of this script, `drjit_path`, is optional and
+indicates the path to the `drjit` package folder if it is not installed or
+included in the user's $PYTHONPATH envvar.
+"""
+
+import os
+import inspect
+import re
+import sys
+import logging
+
+# ------------------------------------------------------------------------------
+
+buffer = ''
+
+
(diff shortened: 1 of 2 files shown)


dvicini:
Hi Christian,

Thanks for the PR. Are you sure that the stub files are necessary for drjit? As far as I remember we added them to Mitsuba because it uses a highly customized module setup. We reroute calls to the Mitsuba module to a specific underlying variant, so there is no way autocomplete tools can work without providing stub files. I thought the situation for drjit was different, as it has a much simpler import mechanism. Is that incorrect?

FYI the stub generation does not seem to work yet on the CI server.

Delio

DoeringChristian:
Hi Delio,

I have tested multiple linters on my system: 
- VSCode with Python extension and Pylance:
   Lists modules deleted at the end of __init__.py (generic, matrix, router)
- VSCode with Jedi:
   Does not list those modules but does not show items assigned with self[k] = v
pylsp and pyright (nvim) seem to have similar problems.
Do you have linting/atuocompletion enabled and does it work for you?
If so it has probably something to do with my setup and I'm sorry for bothering you.

I will however try to fix the code so that it compiles on the CI server.
Sorry for not doing that in beforehand.

Christian

DoeringChristian:
Everything compiles again. I just missed a necessary CMake dependency for it to work.
It did compile on my system because it compiled the targets in a different order.

dvicini:
Okay. I haven't personally tested the autocompletion in VS code recently. I really thought it was working reasonably well out of the box. I will wait with merging this PR until the rest of the team is back from their vacations to take another look. Since this is a mainly "additive" PR, hopefully waiting with the merge doesn't imply much extra work. 

Thanks for your help on this!

njroussel:
Hi @DoeringChristian 

Thank you for this, I've added the last few missing details in #61 if you want to take a look.
The Python stubs will be available directly through `pip install drjit` with the next release, until then you will need to build the project on your machine.

PR #55, opened 2022-08-05
Patch/gcc11

tomix1024:
When including the intrinsics in `include/drjit/packet_intrin.h`, GCC expects us to include the `immintrin.h` header instead of the individual headers.
In GCC 11 however, the `bmi2intrin.h` headers are expected to be included via `x86gprintrin.h`.
This PR adds the missing "workaround" definition if GCC >= 11.

### DIFF
--- a/include/drjit/packet_intrin.h
+++ b/include/drjit/packet_intrin.h
@@ -26,6 +26,8 @@
 #  if !defined(_IMMINTRIN_H_INCLUDED) && !defined(__clang__) && defined(__GNUC__)
      // And the same once more, for GCC..
 #    define _IMMINTRIN_H_INCLUDED
+// GCC 11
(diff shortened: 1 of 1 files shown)


merlinND:
Thanks for the PR @tomix1024 !
Looking at the build log, it looks like the CI failure is unrelated to your change (rather some driver issue on the CI machine).
We'll take a look when we're back from SIGGRAPH.

njroussel:
Sorry, I completely lost track of this.

This looks good to me.

I checked the headers of GCC 11 and 12, we should be able to get away for all our intrinsic includes with just `_IMMINTRIN_H_INCLUDED` and `_X86GPRINTRIN_H_INCLUDED`.

Right now, `_X86INTRIN_H_INCLUDED` is not needed.

Issue #63, opened 2022-08-18
Missing overload for Arm Neon intrinsics


flanggut:
Hi! I am evaluating if it is worth switching some code from enoki to drjit. I noticed that currently there is no packet overload for Neon intrinsics, i.e. something corresponding to enokis [array_neon.h](https://github.com/mitsuba-renderer/enoki/blob/master/include/enoki/array_neon.h). 
Are you planning on adding this at a later point, or would you welcome some PRs adding this back?

Speierers:
In Dr.Jit we swtiched to JIT compiled LLVM kernel that can be executed in parallel on CPU. This sorts of replaces the `packet` mode of Enoki for efficient vectorization.

This being said Dr.Jit still exposes the packets type as we find those convient in some cases. I am not sure onto why we dropped the support for Neon intrinsics. @wjakob will probably know better.

wjakob:
Hi -- this fell under the table at some point because it didn't quite seem important enough given the strong focus on dynamic compilation. We would definitely be interested in a ported version of the NEON intrinsic backend.

PR #70, opened 2022-09-03
feat: neon intrinsics packet backend

flanggut:
Adding overloads for a neon intrinsics backend. The code is a ported version of the backend from the enoki repo. Tested via unit tests on an M1 mac. See also #63. 

I had to add some "hacky" changes to make sure all tests compile and run. I'm happy to incorporate better ways to fix those issues:

1. changing the seed for the random generator in `validate_horizontal` in `test.h` 
This is because `assert_close` doesn't really work for floats if `result_ref` is `0` and `result` is something like `1e-7`. I was hitting exactly that case on my system. Changing the seed doesn't fix it it just avoids it.

2. Adding `<ostream>` to `half.h`, and including half in `packet_neon.h`
This is just so `half` is defined in `packet_neon` when the conversion is instantiated.
`half.h` also includes an operator overload for ostreams so this needs to be defined as well.

### DIFF
--- a/include/drjit/half.h
+++ b/include/drjit/half.h
@@ -13,6 +13,8 @@
 
 #pragma once
 
+#include <ostream>
+
 #include <drjit/array_traits.h>
 #include <drjit/packet_intrin.h>
 
--- a/include/drjit/packet_neon.h
+++ b/include/drjit/packet_neon.h
@@ -1,0 +1,1259 @@
+/*
+    drjit/packet_neon.h -- Packet arrays, ARM NEON specialization
+
+    Dr.Jit is a C++ template library for efficient vectorization and
+    differentiation of numerical kernels on modern processor architectures.
+
+    Copyright (c) 2021 Wenzel Jakob <wenzel.jakob@epfl.ch>
+
+    All rights reserved. Use of this source code is governed by a BSD-style
+    license that can be found in the LICENSE file.
+*/
+
+#pragma once
+
+#include <drjit/half.h>
+
+NAMESPACE_BEGIN(drjit)
+DRJIT_PACKET_DECLARE(16)
+DRJIT_PACKET_DECLARE(12)
+
+NAMESPACE_BEGIN(detail)
+static constexpr uint64_t arm_shuffle_helper_(int i) {
+    if (i == 0)
+        return 0x03020100;
+    else if (i == 1)
+        return 0x07060504;
+    else if (i == 2)
+        return 0x0B0A0908;
+    else
+        return 0x0F0E0D0C;
+}
+NAMESPACE_END(detail)
+
+DRJIT_INLINE uint64x2_t vmvnq_u64(uint64x2_t a) {
+    return vreinterpretq_u64_u32(vmvnq_u32(vreinterpretq_u32_u64(a)));
+}
+
+DRJIT_INLINE int64x2_t vmvnq_s64(int64x2_t a) {
+    return vreinterpretq_s64_s32(vmvnq_s32(vreinterpretq_s32_s64(a)));
+}
+
+/// Partial overload of StaticArrayImpl using ARM NEON intrinsics (single precision).
+template <bool IsMask_, typename Derived_> struct alignas(16)
+    StaticArrayImpl<float, 4, IsMask_, Derived_>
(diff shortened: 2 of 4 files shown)


include/drjit/half.h
```diff
@@ -13,6 +13,8 @@
 
 #pragma once
 
+#include <ostream>
+
```
Speierers, line 17: Please remove this

tests/test.h
```diff
@@ -442,7 +442,7 @@ void validate_horizontal(const std::vector<value_t<T>> &args,
                          value_t<T> (*refFunc)(const T &),
                          const value_t<T> &eps = 0) {
     std::mt19937 gen;
-    std::uniform_int_distribution<> dis(0, (int) args.size()-1);
+    std::uniform_int_distribution<> dis(1, (int) args.size()-1);
```
Speierers, line 445: Is this necessary?

include/drjit/packet_neon.h
```diff
@@ -0,0 +1,1259 @@
...
+#endif
+
+    //! @}
+    // -----------------------------------------------------------------------
+
+    // -----------------------------------------------------------------------
+    //! @{ \name Reinterpreting constructors, mask converters
+    // -----------------------------------------------------------------------
+
+#define ENOKI_REINTERPRET_BOOL(type, target)                   \
```
Speierers, line 572:
```suggestion
#define DRJIT_REINTERPRET_BOOL(type, target)                   \
```

include/drjit/packet_neon.h
```diff
@@ -0,0 +1,1259 @@
...
+    // -----------------------------------------------------------------------
+    //! @{ \name Vertical operations
+    // -----------------------------------------------------------------------
+
+    DRJIT_INLINE Derived add_(Ref a) const { return vaddq_f64(m, a.m); }
+    DRJIT_INLINE Derived sub_(Ref a) const { return vsubq_f64(m, a.m); }
+    DRJIT_INLINE Derived mul_(Ref a) const { return vmulq_f64(m, a.m); }
+    DRJIT_INLINE Derived div_(Ref a) const { return vdivq_f64(m, a.m); }
+
+#if defined(ENOKI_ARM_FMA)
```
Speierers, line 421:
```suggestion
#if defined(DRJIT_ARM_FMA)
```

Speierers (commented):
This PR looks good to me (especially since this was ported directly from the Enoki codebase). 

Just a few minor comments to be addressed before we merge this. 

@wjakob could you remind me the reason for not porting this to Dr.Jit from Enoki during the rewrite?

wjakob:
Hi all -- sorry for the inactivity. The upcoming rewrite (`nanobind_v2` branch) will also have an ARM packet backend. Let's leave this PR open until this branch is merged in case others are looking for this.

include/drjit/packet_neon.h
```diff
@@ -0,0 +1,1259 @@
...
+    All rights reserved. Use of this source code is governed by a BSD-style
+    license that can be found in the LICENSE file.
+*/
+
+#pragma once
+
+#include <drjit/half.h>
+
+NAMESPACE_BEGIN(drjit)
+DRJIT_PACKET_DECLARE(16)
```
jeongseok-meta, line 18: This change was needed to build for android32:

```suggestion	
DRJIT_PACKET_DECLARE_COND(16, enable_if_t<(std::is_same_v<Type, float>)>)
DRJIT_PACKET_DECLARE_COND(16, enable_if_int32_t<Type>)
#if defined(ENOKI_ARM_64)
DRJIT_PACKET_DECLARE_COND(16, enable_if_t<(std::is_same_v<Type, double>)>)
DRJIT_PACKET_DECLARE_COND(16, enable_if_int64_t<Type>)
#endif
```

include/drjit/half.h
jeongseok-meta: This header was needed since `std::ostream` is used at:

https://github.com/mitsuba-renderer/drjit/pull/70/files#diff-f3980386897c4296229cba53ff540bab4a045739a3be2c7137f3b96befb9fa1bL71

njroussel:
The changes on `nanobind_v2` have made their way onto `master`, I think it entirely covers this PR.

PR #74, opened 2022-09-08
Remove additional DLL search directories in Python package init

njroussel:
This PR is somewhat of a followup to this issue: https://github.com/mitsuba-renderer/mitsuba3/issues/228

Currently, `drjit` has a `os.add_dll_directory` statement in its `__init__.py` file. Unfortunately, this functionality is not well supported by Conda on Windows (https://github.com/conda/conda/issues/10897). Moreover, this code is only useful when building **Mitsuba** and serves no purpose when using Dr.Jit on its own.
This PR removes the `os.add_dll_directory` statement.

This PR also refactors the Python stub file generation such that it expected the `drjit` package to already be located on the `PYTHONPATH`. This is not a necessary change, but rather for consistency with the work done in https://github.com/mitsuba-renderer/mitsuba3/pull/245.

### DIFF
--- a/drjit/__init__.py
+++ b/drjit/__init__.py
@@ -4,17 +4,6 @@ import os
 if sys.version_info < (3, 8):
     raise ImportError("Dr.Jit requires Python >= 3.8")
 
-if os.name == 'nt':
-    # Specify DLL search path for windows (no rpath on this platform..)
-    d = __file__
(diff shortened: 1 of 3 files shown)


Speierers (commented):
Looking good to me 👍

PR #82, opened 2022-09-23
Meshgrid: add 3D support on the C++ side

merlinND:
Includes C++ unit tests checked against NumPy.


Speierers (approved):
Thanks @merlinND for porting this feature to C++. The patch looks great to me, please go ahead and merge it if the test suit is green. 🚀

PR #84, opened 2022-10-03
Support list/dict when wrapping function in wrap_ad

thomasw21:
When using `wrap_ad` I noticed that the following would fail if you ran `dr.backward()` afterwards:

```
@dr.wrap_ad(source="drjit", target="torch")
def torch_func(x: List[torch.Tensor]):
    return x[0]
```

Previous assumption supposed that arguments had flattened `torch.Tensor` allowing it to just query `grad` in element of `args` . Therefore, it tried to query the attribute `grad` from a list, instead of trying to querying it from the element of the list.

We propose a recursive solution that should work for nested list/dict

### DIFF
--- a/drjit/router.py
+++ b/drjit/router.py
@@ -5869,20 +5869,21 @@ def wrap_ad(source: str, target: str):
                         grad_out_torch = drjit_to_torch(self.grad_out())
                         grad_out_torch = torch_ensure_shape(grad_out_torch, self.res_torch)
                         _torch.autograd.backward(self.res_torch, grad_out_torch)
-                        args_grad_torch = self.get_grads(self.args_torch)
+
+                        def get_grads(args):
+                            if isinstance(args, _Sequence) and not isinstance(args, str):
+                                return tuple(get_grads(b) for b in args)
+                            elif isinstance(args, _Mapping):
+                                return {k: get_grads(v) for k, v in args.items()}
+                            elif is_torch_tensor(args):
+                                return getattr(args, 'grad', None)
+                            else:
+                                return None
+
+                        args_grad_torch = get_grads(self.args_torch)
                         args_grad = torch_to_drjit(args_grad_torch)
                         self.set_grad_in('args', args_grad)
 
-                    @classmethod
-                    def get_grads(cls, args):
-                        if isinstance(args, _Sequence) and not isinstance(args, str):
-                            return tuple(cls.get_grads(b) for b in args)
-                        elif isinstance(args, _Mapping):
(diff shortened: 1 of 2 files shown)


drjit/router.py
```diff
@@ -5869,10 +5869,19 @@ def backward(self):
...
 
+                    @classmethod
+                    def get_grads(cls, args):
+                        if isinstance(args, _Sequence) and not isinstance(args, str):
+                            return tuple(cls.get_grads(b) for b in args)
+                        elif isinstance(args, _Mapping):
+                            return {k: cls.get_grads(v) for k, v in args.items()}
+                        else:
+                            return getattr(args, 'grad', None)
+
```
Speierers, line 5884: As this is only used in one place, it is maybe unnecessary to put this code in a separate method.

Speierers (commented):
Thanks for looking into this. It would be great to add a couple unit-tests and update the docstring if necessary.

Also, do we need similar changes on the Torch side? (e.g. when using `PyTorch` custom operation).

drjit/router.py
thomasw21: It's recursive so it needs it's own function declaration. Do you want it inside the other function?

thomasw21:
> It would be great to add a couple unit-tests and update the docstring if necessary.

Current docstring says: `The wrapped function's arguments will be casted to the tensor types corresponding to the ``target`` framework` which I'm guessing is fine?

> Also, do we need similar changes on the Torch side? (e.g. when using PyTorch custom operation).

Do you mean if the target is `drjit`? I'm a bit new to how things work for `drjit` so I'm not sure, but I don't see such assumption in drjit part, since `drjit_to_torch` seem to be able to handle dict/list

thomasw21:
Btw I'm wondering if there's a simple way to install it with python bindings. What I currently did to install it was:
```
# Run the https://drjit.readthedocs.io/en/latest/firststeps-cpp.html
$ PYTHONPATH="path_to_drjit_repo/build/drjit" pytest text23D/drjit/tests/python/test_wrap_ad.py
```
But I couldn't figure out how to install it through `pip` ... Maybe I'm missing something stupid.

drjit/router.py
Speierers: Right, then I would prefer having it within the other function.

thomasw21:
> But I couldn't figure out how to install it through pip ... Maybe I'm missing something stupid.

Nevermind, actually adding `package_dir={"": "build"}` made it work

Speierers:
Let's merge this 🚀 Thanks for this PR @thomasw21!

PR #88, opened 2022-10-08
Allows functions in `wrap_ad` context to return nested structures

thomasw21:
`torch.autograd.backward` only supports list of tensors: https://pytorch.org/docs/stable/generated/torch.autograd.backward.html . This causes the previous assumption to break as `self.res_torch` could be a nested structure.

We propose to recursively flatten eveything to be passed to the backward step. We also add a test demonstrating that it works.

### DIFF
--- a/drjit/router.py
+++ b/drjit/router.py
@@ -5868,7 +5868,16 @@ def wrap_ad(source: str, target: str):
                     def backward(self):
                         grad_out_torch = drjit_to_torch(self.grad_out())
                         grad_out_torch = torch_ensure_shape(grad_out_torch, self.res_torch)
-                        _torch.autograd.backward(self.res_torch, grad_out_torch)
+                        def flatten(values) -> Union[torch.Tensor, _Sequence[torch.Tensor]]:
+                            """Flatten structure in a consistent arbitrary order"""
+                            if isinstance(values, _Sequence):
+                                return [flatten(v) for v in values]
+                            elif isinstance(values, _Mapping):
+                                return [flatten(v) for k, v in sorted(values.items(), key=lambda k, v: k)]
+                            else:
+                                return values
+
+                        _torch.autograd.backward(flatten(self.res_torch), flatten(grad_out_torch))
 
                         def get_grads(args):
                             if isinstance(args, _Sequence) and not isinstance(args, str):
--- a/drjit/traits.py
+++ b/drjit/traits.py
@@ -813,7 +813,7 @@ def leaf_array_t(arg):
             if is_diff_v(t) and is_float_v(t):
                 break
     elif isinstance(arg, _Mapping):
-        for k, v in arg:
+        for k, v in arg.items():
             t = leaf_array_t(v)
             if is_diff_v(t) and is_float_v(t):
                 break
--- a/tests/python/test_wrap_ad.py
+++ b/tests/python/test_wrap_ad.py
@@ -193,3 +193,27 @@ def test09_to_torch_list_of_tensors_as_args(m):
     assert dr.allclose(dr.grad(a), [4, 4, 4])
     assert dr.allclose(dr.grad(b), [3, 3, 3])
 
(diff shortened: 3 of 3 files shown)


setup.py
```diff
@@ -52,6 +52,7 @@
         f'-DCMAKE_TOOLCHAIN_FILE={drjit_cmake_toolchain_file}',
         f'-DDRJIT_PYTHON_STUBS_DIR:STRING={drjit_python_stubs_dir}'
     ],
+    package_dir={'': 'build'},
```
thomasw21, line 55: @Speierers I added this in order to run pip install after building.

njroussel: This is only necessary if you want to re-use the current build, right? Without this line, it would re-build the entire project.

FWIW Dr.Jit  supports in-tree builds, having an external build folder and naming it `build`  is not mandatory. Also, it's much less error-prone to re-build the entire project that re-using an existing build.
I wouldn't add this line unless there is something less.

thomasw21: Hum I couldn't seem to install it otherwise ... Not sure what was happening. I'll remove it.

njroussel: Oh, that's surprising. Can you try running `pip install ../drjit` from the drjit's root directory?

thomasw21: Seems to work now. My bad then

drjit/traits.py
```diff
@@ -813,7 +813,7 @@ def leaf_array_t(arg):
             if is_diff_v(t) and is_float_v(t):
                 break
     elif isinstance(arg, _Mapping):
-        for k, v in arg:
+        for k, v in arg.items():
```
Speierers, line 816: Nice catch!

drjit/router.py
```diff
@@ -5868,7 +5868,27 @@ def forward(self):
...
+                                    keys = input.keys()
+                                for key in keys:
+                                    first = input[key]
+                                    second = other[key]
+                                    flattened_elt = flatten(first, second)
+                                    flattened_input += flattened_elt[0]
+                                    flattened_output += flattened_elt[1]
+                                return flattened_input, flattened_output
+
+                        _torch.autograd.backward(*flatten(self.res_torch, grad_out_torch))
```
Speierers, line 5891: I would rewrite this as follow (untested):

```python
def flatten(values):
    if (isinstance(values, _Sequence):
        return [flatten(v) for v in values]
    elif isinstance(values, _Mapping)):
        return [flatten(v) for k, v in values.item()]
    else:
        return values

_torch.autograd.backward(flatten(self.res_torch), flatten(grad_out_torch))
```

Speierers (commented):
Please take a look at the proposed implementation. If this works, then you can squash the commits and I will take a last look before merging this PR. 🚀

drjit/router.py
thomasw21: Hum I thought about it, but you need the structure to match in their flattened ordering. So typically `.items()` doesn't guarantee you ordering, so if you process `self.res_torch` and `grad_out_torch` seperately then you might not have aligne structures anymore.

drjit/traits.py
thomasw21: TBH, it got caught when I ran things on my side.

drjit/router.py
Speierers: In this case could we simply cast those into `OrderedDict` before calling `.items()`?

thomasw21: This could work, used `sorted` instead in 9f2db3bdd04fc8ba61409f74a23592193e8eb5e6 . Tests seem to pass.

Speierers:
Great! Please squash those commits into a single one and then we can merge this PR.

thomasw21:
I can do this. Though GH usually should have an option to prevent merge commits in settings. I usually have the following deactivated.
image

Speierers:
Thanks for your contribution! 🚀

PR #95, opened 2022-11-09
Fix Python type checking in wrap_ad

njroussel:
Fixes failing tests because of invalid Python type hints.


Speierers (commented):
Nice catch! 🚀

PR #98, opened 2022-11-17
Mask stack improvements

wjakob:
This commit contains the Dr.Jit part of two related PRs (with the Dr.Jit-Core part being at https://github.com/mitsuba-renderer/drjit-core/pull/32)

It bundles several improvements to the mask stack. For context, the mask stack is an implicit stack of masks, whose top element is automatically applied to memory reads/writes and a few other operations to disable inactive entries of the current wavefront. That is important because undefined behavior resulting from unmasked memory operations could cause inconsistencies or even crash the program.

LLVM has an implicit default mask (even when the stack is empty) that is needed because the wavefront size might not be an exact multiple of the SIMD vector width. This means that some elements are always masked even in the absence of conditional operations that normally cause some masking (loops, vcalls).

The changes are:

- The implicit default mask on LLVM is now consistently created based on the size of the underlying wavefront. That requires changes in a few places because the size has to be determined. Previously, the default mask always had a size of 1, which made no sense and could cause serious issues when somebody actually evaluated the mask and turned into a memory-backed variable.

- Several parts of Dr.Jit fetch the top of the mask stack and combine it with an existing mask if the sizes are compatible. The commit adds a function `jit_var_mask_apply()` to factor out this repeated code.

- The ``dr::Loop::operator()`` routine was refactored.

- The `jit_var_vcall()` routine for recorded dynamic dispatch now incorporates the `jit_var_mask_apply()` logic as well as a check for whether `self == nullptr`. That means that these steps can be removed from the caller.

### DIFF
--- a/include/drjit/vcall_jit_record.h
+++ b/include/drjit/vcall_jit_record.h
@@ -123,7 +123,7 @@ Result vcall_jit_record_impl(const char *name, uint32_t n_inst,
                 nullptr));
         }
 
-        jit_state.set_mask(vcall_mask.index(), false);
+        jit_state.set_mask(vcall_mask.index());
 
         if constexpr (std::is_same_v<Result, std::nullptr_t>) {
             func(base, (set_mask_true<Is, N>(args))...);
@@ -216,7 +216,7 @@ Result vcall_jit_record(const char *name, const Func &func, Self &self,
     auto [inst, n_inst] = vcall_registry_get(Backend, Base::Domain);
 
     size_t self_size = width(self, args...);
-    Mask mask = extract_mask<Mask>(args...) && neq(self, nullptr);
+    Mask mask = extract_mask<Mask>(args...);
     bool masked = mask.is_literal() && mask[0] == false,
          vcall_inline = jit_flag(JitFlag::VCallInline);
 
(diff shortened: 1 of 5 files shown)


include/drjit/vcall_jit_record.h
```diff
@@ -232,15 +232,13 @@ Result vcall_jit_record(const char *name, const Func &func, Self &self,
...
+            (Base *) inst, func, mask && neq(self, nullptr),
             std::make_index_sequence<sizeof...(Args)>(), args...);
     } else {
-        // Also check the mask stack to constrain side effects in recorded computation
-        Mask mask_combined = mask && Mask::steal(jit_var_mask_peek(Backend));
         isolate_grad<DiffType> guard;
 
         return vcall_jit_record_impl<Result, Base>(
-            name, n_inst, func, self, mask_combined,
+            name, n_inst, func, self, mask,
```
Speierers, line 241: Don't we need the ` && neq(self, nullptr)` here as well?

Speierers (commented):
LGTM, one minor comment.

include/drjit/vcall_jit_record.h
wjakob: That check is now done inside of `jit_var_vcall`. We only need it for the scalar case. This, btw, might also impact your Python wrapper for the vcalls.

wjakob:
I merged the drjit-core part and rebased this PR.

Issue #43, opened 2022-07-20
Unexpected behavior of drjit.shape


JamesZFS:
I am playing with drjit in python and seem to find an undesirable behavior of the `shape` function in `router.py`:
If you define a ragged 2D array whose # entries of the second dimension decrease, the output of `dr.shape` is None, which is the expected result of ragged arrays.
```python
a = mi.Color3f([0, 1, 2], [0, 1], [0, 1])
dr.shape(a)
```
out:
```
None
```
But if you define it in a way where # entries of the second dimension increase, the output is not None, but instead behave like the array is extended to a uniform-length array
```python
a = mi.Color3f([0, 1], [0, 1, 2], [0, 1, 2, 3])
dr.shape(a)
```
out:
```
[3, 4]
```

Speierers:
This will be fixed in the next release: a026b56

PR #100, opened 2022-11-25
Implement `dr.switch()` and `dr.dispatch()`

Speierers:
This PR implements `dr.switch()` and `dr.dispatch()`


drjit/router.py
```diff
@@ -5947,3 +5945,205 @@ def get_cmake_dir():
...
+        funcs (list): a list of functions to dispatch based on indices
+        args (tuple): the arguments to pass to the functions
+
+    Returns:
+        object: the result of the function call dispatched based on the indices
+    """
+    if not _dr.flag(_dr.JitFlag.VCallRecord):
+        raise Exception("switch(): only supports recorded function dispatch. "
+                        "To set the correct flag, use the following command: "
+                        "dr.flag(dr.JitFlag.VCallRecord, True).")
```
njroussel, line 5987: Should we have an implementation without recording?

For example, an integrator written with `dr.dispatch` would have to have a big if/else statement if the user wishes to disable recording for debugging purposes. I can see this being an annoyance.

drjit/router.py
```diff
@@ -5947,3 +5945,205 @@ def get_cmake_dir():
...
+    def generate(inst):
+        return (lambda *args: func(inst, *args)) if inst else None
+
+    n_inst_max = Type.registry_get_max_()
+    funcs = [generate(Type.registry_get_ptr_(i+1)) for i in range(n_inst_max)]
+
+    index_type = getattr(_modules.get(Type.__module__), 'UInt32')
+    indices = _dr.reinterpret_array_v(index_type, instances)
+
+    return _dr.switch(indices-1, funcs, *args)
```
njroussel, line 6148: Why is there a `-1` here ?

Speierers: Because the indices for registered instances start at `1` and `dr.switch` expects it to start at `0`

njroussel (approved):
LGTM, don't forget to push the few `drjit-core` changes you made too :)

wjakob:
> don't forget to push the few drjit-core changes you made too :)

Curious about those, I do not see them.

Speierers:
This PR now relies on a few changes in `drjit-core` -> [PR #43](https://github.com/mitsuba-renderer/drjit-core/pull/43)

drjit/router.py
```diff
@@ -5947,3 +5945,196 @@ def get_cmake_dir():
...
+
+
+def dispatch(instances, func, *args):
+    """
+    Dispatches a call to the given function on an array of instances of a class
+    registered to the Dr.Jit registry. The provided function is assumed to take
+    at least one argument that will represent the individual instance of the
+    class this function is recorded on.
+
+    This routine can be used to emulate virtual function dispatching directly
```
wjakob, line 6093: "This routine can be used to perform dynamic dispatch to Python methods of `instances`. The code generation underlying Dr.Jit will transform arbitrary sequences of such method calls into unified functions in the generated code. In other words, a single dispatch call that calls two methods is potentially significantly faster than two separate dispatches."

drjit/router.py
```diff
@@ -5947,3 +5945,196 @@ def get_cmake_dir():
...
+            v = self.foo() # call foo() on a single instance at a time
+            return v + arg
+
+        arg = dr.llvm.Float([1.0, 2.0, 3.0, 4.0])
+
+        res = dr.dispatch(pointers, func, arg)
+
+    .. danger::
+
+        This routine only works with the flag VCallRecord enabled.
```
wjakob, line 6110: no longer true, right?

tests/python/test_switch.py
```diff
@@ -0,0 +1,100 @@
+import drjit as dr
+import pytest
+
+
+def get_module(name):
```
wjakob, line 5: Tests for failure modes are needed. Can we have a non-differentiable and a differentiable dispatch call that each fail with an exception? Such things are expected to happen when overriding Python scene objects from a Jupyter notebook session.

src/python/switch.h
```diff
@@ -0,0 +1,177 @@
...
+                              py::args args) {
+    static constexpr JitBackend Backend = backend_v<UInt32>;
+    using Mask = dr::mask_t<UInt32>;
+    using DiffType = leaf_array_t<dr::float_array_t<UInt32>>;
+
+    py::object detail = py::module_::import("drjit").attr("detail");
+    auto apply_cpp = detail.attr("apply_cpp");
+
+    // Schedule arguments, will be evaluated in jit_var_vcall_reduce
+    apply_cpp(args, py::cpp_function([&](uint32_t index) {
```
wjakob, line 129: is the `&` needed here?

src/python/switch.h
```diff
@@ -0,0 +1,177 @@
...
+        size_t wavefront_size = perm.size();
+
+        MaskScope<Mask> scope(Mask::steal(
+            jit_var_mask_default(Backend, (uint32_t) wavefront_size)));
+
+        // Avoid merging multiple vcall launches if size repeats..
+        if (wavefront_size != last_size) {
+            last_size = wavefront_size;
+        } else {
+            apply_cpp(result, py::cpp_function([&](uint32_t index) {
```
wjakob, line 157: is the `&` needed here?

wjakob (commented):
Looks great Sébastien -- some feedback from me.

PR #102, opened 2022-11-29
Adapt Dr.Jit to Dr.Jit-Core code generation

wjakob:
This PR makes Dr.Jit compatible with the following Dr.Jit-Core PRs:

- https://github.com/mitsuba-renderer/drjit-core/pull/40
- https://github.com/mitsuba-renderer/drjit-core/pull/41

### DIFF
--- a/include/drjit/vcall_jit_record.h
+++ b/include/drjit/vcall_jit_record.h
@@ -96,7 +96,6 @@ Result vcall_jit_record_impl(const char *name, uint32_t n_inst,
 
     detail::JitState<Backend> jit_state;
     jit_state.begin_recording();
-    jit_state.new_scope();
 
     state[0] = jit_record_checkpoint(Backend);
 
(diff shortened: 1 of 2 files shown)


include/drjit/vcall_jit_record.h
```diff
@@ -236,6 +236,7 @@ Result vcall_jit_record(const char *name, const Func &func, Self &self,
             std::make_index_sequence<sizeof...(Args)>(), args...);
     } else {
         isolate_grad<DiffType> guard;
+        jit_new_scope(Backend);
```
Speierers, line 239: Do we really need those two scopes here?

Speierers (commented):
One question, otherwise LGTM

njroussel (approved):

include/drjit/vcall_jit_record.h
wjakob: Here, the idea was to wrap all of the vcalls into their own scopes so that LVN doesn't merge things between caller and callee. It's possible that it could work without this, but I don't think there is any harm.

PR #106, opened 2022-12-07
Fix vcall scoping when symbolic loop in vcalls

Speierers:
This patch forces the JIT scope to be set between the tracing of each VCall to deal with the case of symbolic loops within vcalls.


wjakob:
LGTM!

PR #133, opened 2023-03-20
Improve testing of conversions to other frameworks

dvicini:
This PR makes sure we test conversion to and from other frameworks a bit more exhaustively, in particular for LLVM/CPU modes. The changes are simple, but I am making a PR before merging to see if it passes the CI

### DIFF
--- a/tests/python/test_conversions.py
+++ b/tests/python/test_conversions.py
@@ -52,10 +52,12 @@ def test_roundtrip_numpy_all(package):
         assert len(Float(np.array([], dtype=np.float32))) == 0
 
 
-def test_roundtrip_pytorch_cuda():
+@pytest.mark.parametrize("package", ["drjit.cuda", "drjit.cuda.ad",
+                                     "drjit.llvm", "drjit.llvm.ad"])
+def test_roundtrip_pytorch_all(package):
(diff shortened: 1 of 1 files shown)


dvicini:
That's a good point, so the tests will either way not cover the full set of permutations. I don't have a local GPU, so I was surprised to see that some of these tests were requiring a GPU to be run. I think it makes sense for these tests to also execute when no CUDA GPU is available. If you know a better way of doing this, I am happy to change the PR

njroussel:
Quick update:

I've been considering changing the behaviour of `mi.variants()`. Currently, it will show backends that are compiled and not backends that are actually available on the user's system. Of course, this could also be done with a new function, it doesn't have to be a change to `mi.variants()`.

This should simplify some of the testing framework and will also solve https://github.com/mitsuba-renderer/mitsuba3/issues/516.

This is still not a priority, but I want to work on it in the next few months - I've been using a setup without a local GPU and this has become very annoying.

wjakob:
The nanobind_v2 branch does quite a lot of tests for these kinds of conversions. I am thinking that this commit will likely be superseded by the release of those changes.

dvicini:
Closing this as the new tests indeed supersede those changes

PR #135, opened 2023-03-23
Fix array shape in dr.quat_to_matrix

tomix1024:
If I have a simple `mi.Quaternion4f q` in Python, the matrix type is not correctly derived when calling `dr.quat_to_matrix(q)`.

### DIFF
--- a/drjit/matrix.py
+++ b/drjit/matrix.py
@@ -339,7 +339,7 @@ def matrix_to_quat(m, /):
     if not _dr.is_matrix_v(m):
         raise Exception('Unsupported type!')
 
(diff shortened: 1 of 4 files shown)


njroussel:
Hi @tomix1024 

Thank you, this indeed seems correct. Do you mind adding a test for it too in `test_matrix.py`?

tomix1024:
Hi @njroussel,

That's actually a good idea, I've added some tests.
There was also a problem with `dr.select(mi.Bool, mi.Quaternion4f, mi.Quaternion4f)`, which tried to create a value of type `mi.Quaternion4b` internally, which does not exist...

njroussel:
Great, I don't have anything to add. 
Thank you very much!

Issue #85, opened 2022-10-06
Bug: tensor multi index broadcasting fails.


Jerry-Shen0527:
I tried to do multi index tensor broadcasting and the indices go out of range. I went through the unit tests and found only single index broadcasting, so I guess multi index tensor broadcasting has not been tested.

![image](https://user-images.githubusercontent.com/47278074/194262217-598ec0c3-3149-4243-81fb-1a40cc77ca99.png)

I'm currently adding `else` before `size = size_next` in file tensor.h and this temporarily solves the problem.

Speierers:
Hi, 

To better understand your issue, could you please provide a simple reproducer code?

Jerry-Shen0527:
I tried to print out the indices used by gather in `tensor_broadcast_impl()` and in such multi indexed cases, some of the indices would go out of bounds.

I don't know whether this means I should not do multi indexed broadcasting, or it's just a miscalculation of the index.

Speierers:
Hi @Jerry-Shen0527, indeed multiple-index broadcasting is likely not properly supported in Dr.Jit. It would be great to at least throw an exception when such a case is detected.

I will add this on our TODO list. For now I would recommend you stick with single index broadcasting or implement a fix yourself (in which case a PR would be greatly appreciated).

njroussel:
Fixed with #146

PR #146, opened 2023-05-07
fix bug in tensor broadcasting

Andy3531:
The original code will fail on multipling two tensors of shape (512,512,4) and (1,1,4) because of the wrong recursive direction in the for loop.

### DIFF
--- a/include/drjit/tensor.h
+++ b/include/drjit/tensor.h
@@ -30,12 +30,11 @@ void tensor_broadcast_impl(const char *op, T &t, const dr_vector<size_t> &shape)
         size *= (uint32_t) shape[i];
 
     Index index = arange<Index>(size);
-    size = 1;
 
-    for (int i = ndim - 1; i >= 0; --i) {
-        uint32_t size_next = size * (uint32_t) shape[i];
+    for (int i = 0; i < ndim; ++i) {
+        uint32_t size_next = size / (uint32_t) shape[i];
         if (t.shape(i) == 1 && shape[i] != 1)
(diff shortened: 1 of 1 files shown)


njroussel:
Hi @Andy3531

Could you provide us with a reproducer for the original issue? I tried something naive and it seems to be working as intended.

Andy3531:
Yes, if we run the following code,
```
size_t shape0[3]={512,512,4}; 
TensorXf tensor0(dr::zeros<typename TensorXf::Array>(512*512*4), 3, shape0);`

size_t shape1[3]={1,1,4};
TensorXf tensor1(dr::zeros<typename TensorXf::Array>(4), 3, shape1);

tensor0 = tensor0.mul_(tensor1);
```
we will get an error: **Caught a critical exception: ArrayBase: out of range access (tried to access index 4 in an array of size 4)**

where type `TensorXf` is defined here

https://github.com/mitsuba-renderer/mitsuba3/blob/9273af4798f27a9235fc275753a2779088fb4bfd/include/mitsuba/core/fwd.h#L179

Andy3531:
And similar problem is encountered in this issue:
@Speierers in https://github.com/mitsuba-renderer/drjit/issues/85#issuecomment-1272843300_

njroussel:
It took me a while to figure it out, but I finally was able to reproduce this...
So, in JIT variants this doesn't crash, but produces a potentially wrong output. Only in scalar mode does the error message appear.

njroussel (approved):
The new indexing is correct, thank you!

PR #154, opened 2023-05-21
Fix implicit gradient tracking in dr.switch and dr.dispatch

Speierers:
This patch fixes various issues with `dr.switch` and `dr.dispatch` when the functions have an implicit dependency to variables that have gradient tracking enabled. This was not previously tested, but it is a crucial functionality that needs to be incorporated. Additionally, this PR adds the related unit tests.

### DIFF
--- a/src/python/switch.h
+++ b/src/python/switch.h
@@ -104,7 +104,7 @@ py::object switch_record_impl(UInt32 indices, py::list funcs, py::args args) {
     uint32_t offset = 0;
     return apply_cpp(result, py::cpp_function([&](uint32_t /*index*/) {
                          return indices_out[offset++];
-                     }));
+                     }), false);
 }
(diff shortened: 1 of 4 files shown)


src/python/switch.h
```diff
@@ -104,7 +104,7 @@ py::object switch_record_impl(UInt32 indices, py::list funcs, py::args args) {
     uint32_t offset = 0;
     return apply_cpp(result, py::cpp_function([&](uint32_t /*index*/) {
                          return indices_out[offset++];
-                     }));
+                     }), false);
 }
```
njroussel, line 108: It took me a while to understand why this was needed. I don't really have any suggestions to improve though.

njroussel (approved):
Thanks for the fix :100:

PR #172, opened 2023-07-14
Fix `dr.backward_from`/`dr.foward_from` on special types

njroussel:
This PR fixes `dr.backward_from` and `dr.formward_from` when used on special types (`IsSpecial == True`).

The current behavior of these functions is to set the gradients to 1 and then propagate them. This first step is complex for certain types, as a gradient of one should be set to all components of the type. This is currently not the case, this PR applies the necessary changes.

Note: This change is specific to `backward_from` and `forward_from`. It seems tempting to fix this in `set_grad` but this could lead to misleading behaviors. For example, let's consider the case of setting the gradient of a quaternion with a single scalar value. We would want to preserve the following equivalence: `dr.set_grad(quat, 1) := dr.set_grad(quat, mi.Quaternion4f(1))`. Neither sides of this equality is setting the gradient of the equation to 1 for all of its components which is not desired behavior for `backward_from`/`forward_from`.

### DIFF
--- a/drjit/router.py
+++ b/drjit/router.py
@@ -4383,7 +4383,7 @@ def forward_from(arg, flags=_dr.ADFlag.Default):
     '''
     ta = type(arg)
     _check_grad_enabled('forward_from', ta, arg)
-    set_grad(arg, 1)
+    set_grad(arg, _dr.ones(ta))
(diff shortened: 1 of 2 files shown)


rtabbara (approved):
lgtm!

PR #183, opened 2023-08-23
Fix various bugs in dr.switch and dr.dispatch

Speierers:
This PR addresses several issues in the `dr.switch` and `dr.dispatch` routines. Key improvements include:

* `dr.grad` now correctly handles mask types.
* It fixes a missing call to `ad_copy` in the switch vcall custom op, which was causing differentiation issues with `dr.switch`. To verify this fix, a test case was added.
* An error is thrown when return types are not consistent, making it easier to identify this problem.


njroussel (approved):
Took me a while to understand this edge case :exploding_head: 
Nice catch!

PR #186, opened 2023-08-31
Fix gather implementation for custom data-structures

Speierers:
This PR fixes compilation issues arising when using the `dr::gather` operation on a custom Dr.Jit data-structure. The previous trait used to compute the mask type couldn't work properly with data structures that contain JIT arrays.


njroussel (approved):
LGTM ! :+1:

PR #187, opened 2023-09-12
Missing const modifier in block_sum functions

mcrescas:
This PR just adds the `const` modifier to the definitions of the function `block_sum` inside `jit.h` and `autodiff.h` that were missing previously.


njroussel (approved):
LGTM :+1:

Issue #188, opened 2023-09-12
Missing definition of mulhi_scalar()


jeongseok-meta:
The function `mulhi_scalar()` is called within `array_utils.h`, but it seems not defined in drjit:

https://github.com/mitsuba-renderer/drjit/blob/c71ebb792c88abf2460e296e213cbcb8069388a8/include/drjit/array_utils.h#L293

which was previously defined in Enoki's [array_fallbacks.h](https://github.com/mitsuba-renderer/enoki/blob/2a18afa2402e0677887c8439fa3d6a270ea15726/include/enoki/array_fallbacks.h#l339-L383).

Is the omission of this function's definition in drjit intentional? If so, could you provide insights on how to address calls to this function? If not, are there plans to integrate this function into the drjit repository?

njroussel:
Fixed in #189

PR #189, opened 2023-09-12
Add missing mulhi_scalar()

jeongseok-meta:
This PR addresses #188 that `mulhi_scalar()` is missing when it's called within `array_utils.h`:

https://github.com/mitsuba-renderer/drjit/blob/c71ebb792c88abf2460e296e213cbcb8069388a8/include/drjit/array_utils.h#L293

which was previously defined in Enoki's [array_fallbacks.h](https://github.com/mitsuba-renderer/enoki/blob/2a18afa2402e0677887c8439fa3d6a270ea15726/include/enoki/array_fallbacks.h#l339-L383).

### DIFF
--- a/include/drjit/array_utils.h
+++ b/include/drjit/array_utils.h
@@ -290,7 +290,7 @@ DRJIT_INLINE T mulhi_(T x, T y) {
         if constexpr (std::is_signed_v<T>) {
             int32_t x1 = (int32_t) (x >> 32);
             int32_t y1 = (int32_t) (y >> 32);
-            uint32_t x0y0_hi = mulhi_scalar(x0, y0);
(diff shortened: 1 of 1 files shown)


njroussel:
Hi @jeongseok-meta 

Indeed, this seems like a code path that was lost to time.

I am confused with your suggestion. Wouldn't it be enough to replace the call to `mulhi_scalar` to `mulhi`? Your implementation does the same thing. I feel like I'm missing a small but important detail.

jeongseok-meta:
Yes, you're right. I've replaced the call from `mulhi_scalar` to `mulhi` for clarity. Thanks for pointing it out.

njroussel (approved):
Great, thanks :+1:

PR #190, opened 2023-09-14
Fix and improve dr.wrap_ad

Speierers:
This pull request fixes an issue with the `wrap_ad` function when dealing with integer PyTorch tensor arguments. The internal routine responsible for reshaping the gradient tensors was failing as integer tensors don't produce any gradients. 

This fix has been tested and verified to work correctly.

Moreover, this PR adds the ability to specify the `source` and `target` directly passing the module. Here is an example:

```python
import torch
import drjit as dr

dr.wrap_ad(source=torch, target=dr)
def my_func(args):
    ...
```

### DIFF
--- a/drjit/router.py
+++ b/drjit/router.py
@@ -5781,13 +5781,13 @@ def wrap_ad(source: str, target: str):
                 else:
                     return a
 
-            # Ensure tensors in `a` have same shape as tensors in `b` (handles dim==0 case)
-            def torch_ensure_shape(a, b):
+            # Ensure gradient tensors in `a` have same shape as tensors in `b` (handles dim==0 case)
+            def torch_ensure_grad_shape(a, b):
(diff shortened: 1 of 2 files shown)


drjit/router.py
```diff
@@ -5722,12 +5722,16 @@ def torch_func(x):
         Forward-mode AD isn't currently supported by this operation.
 
     Args:
-        source (str): The AD framework used outside of the wrapped function.
-        target (str): The AD framework used within the wrapped function.
+        source (str, module): The AD framework used outside of the wrapped function.
+        target (str, module): The AD framework used within the wrapped function.
```
njroussel, line 5726:
```suggestion
        source (str | module): The AD framework used outside of the wrapped function.
        target (str | module): The AD framework used within the wrapped function.
```

njroussel (approved):
All good :100:

PR #191, opened 2023-09-15
Expanding tzcnt() compatibility for non-BMI platforms

jeongseok-meta:
To further accommodate diverse platforms and compilers, we would like to use the `tzcnt()` function even when BMI is not supported. However, this can lead to compilation errors:

```
drjit/include/drjit/packet_intrin.h:304:20: error: use of undeclared identifier '_tzcnt_u32'
        return (T) _tzcnt_u32((unsigned int) v);
                   ^
drjit/include/drjit/packet_intrin.h:307:20: error: use of undeclared identifier '_tzcnt_u64'
        return (T) _tzcnt_u64((unsigned long long) v);
                   ^
2 errors generated.
```

This is because [`_tzcnt_u32/u64` are defined with BMI support](https://en.wikipedia.org/wiki/X86_Bit_manipulation_instruction_set).

This PR introduces a fallback for such scenarios, currently catering to GCC and Clang compilers.


njroussel (approved):
Looks good to me! 

Thanks :100:

PR #200, opened 2023-11-22
Integration of FP16 support and exposing texture Python bindings

rtabbara:
Also:
* Port `texture.cpp` tests to Python. These tests now also run for other variants besides just CUDA
* Added some ref docs for textures

Note, there are other `drjit-core` changes required to support this aside from the original `fp16` [PR](https://github.com/mitsuba-renderer/drjit-core/pull/70). Another drjit-core PR will be put up once the first one has been merged into the nanobind branch.


docs/reference.rst
```diff
@@ -861,6 +864,18 @@ Scalars
     :show-inheritance:
 .. autoclass:: drjit.scalar.ArrayXb
     :show-inheritance:
+.. autoclass:: drjit.scalar.Array0f16
```
wjakob, line 867: Is this code rebased onto the latest version? IIRC I moved most of these array declarations from ``reference.rst`` into a separate file named ``array_types.rst``

include/drjit/array_router.h
```diff
@@ -1045,7 +1046,7 @@ Target gather(Source &&source, const Index &index, const Mask &mask_ = true,
         return result;
     } else {
         /// Case 4: gather<float>(const float *, ...)
-        static_assert(std::is_integral_v<Index> && std::is_scalar_v<Target>,
+        static_assert(drjit::is_array_integral_v<Index> && drjit::is_scalar_v<Target>,
```
wjakob, line 1049: Could you rename `is_array_integral` to `is_integral_array`?

include/drjit/array_router.h
```diff
@@ -1549,7 +1550,10 @@ template <typename T> T replace_grad(const T &a, const T &b) {
                 drjit_raise("replace_grad(): internal error!");
         }
 
-        return T::create_borrow(vb.index_ad(), va.detach_());
+        uint64_t combined_index = ((uint32_t) a.index_combined())
```
wjakob, line 1553: This seems dangerous -- if the input arrays both have gradients attached, you'll blend their AD indices. How about
```
return T::borrow(a.index() | (((uint64_t) b.index_ad()) << 32));
```

include/drjit/autodiff.h
```diff
@@ -588,6 +588,12 @@ struct DRJIT_TRIVIAL_ABI DiffArray
 
     size_t size() const { return jit_var_size(m_index); }
 
+    void resize(size_t size) {
```
wjakob, line 591: This operation seems problematic. It forgets about the AD handle and replaces the variable contents with a resized version of the Dr.Jit variable. How/why is this needed?

src/extra/autodiff.cpp
```diff
@@ -1012,14 +1020,21 @@ JitIndex ad_grad(Index index) {
...
 
-    if (type != VarType::Float32 && type != VarType::Float64)
-        ad_raise("ad_grad(): this is not a floating point variable!");
-
-
-    if (!result.valid())
-        result =
-            JitVar::steal(type == VarType::Float32 ? jit_var_f32(backend, 0.f)
-                                                   : jit_var_f64(backend, 0.0));
+    if (!result.valid()) {
```
wjakob, line 1023: Let's simplify this whole switch statement to.
```
uint64_t zero = 0;
result = JitVar::steal(jit_var_literal(backend, type, &value, 1));
```

src/extra/autodiff.cpp
```diff
@@ -1697,11 +1712,22 @@ struct Gather : Special {
             return;
         }
 
-        if (!source_grad.valid())
-            source_grad =
-                JitVar::steal((VarType) source->type == VarType::Float32
-                                  ? jit_var_f32(backend, 0.f)
-                                  : jit_var_f64(backend, 0.0));
+        if (!source_grad.valid()) {
+            VarType type = (VarType)source->type;
```
wjakob, line 1716: Let's simplify this whole switch statement to.
```
uint64_t zero = 0;
result = JitVar::steal(jit_var_literal(backend, type, &value, 1));
```

src/extra/autodiff.cpp
```diff
@@ -1746,11 +1772,22 @@ struct Scatter : Special {
...
-            target_grad =
-                JitVar::steal((VarType) target->type == VarType::Float32
-                                  ? jit_var_f32(backend, 0.f)
-                                  : jit_var_f64(backend, 0.0));
+        if (!target_grad.valid()) {
+            VarType type = (VarType) target->type;
+            switch (type) {
+                case VarType::Float16:
+                    target_grad = JitVar::steal(jit_var_f16(backend, drjit::half(0)));
+                    break;
```
wjakob, line 1780: Let's simplify this whole switch statement to.
```
uint64_t zero = 0;
result = JitVar::steal(jit_var_literal(backend, type, &value, 1));
```

src/extra/autodiff.cpp
wjakob: Edit: given that it's needed by so many places, perhaps we should just make a new function `jit_var_zero(backend, type)`?

src/extra/autodiff.cpp
wjakob: Edit: given that it's needed by so many places, perhaps we should just make a new function `jit_var_zero(backend, type)`?

src/extra/autodiff.cpp
wjakob: Edit: given that it's needed by so many places, perhaps we should just make a new function `jit_var_zero(backend, type)`?

src/python/docstr.h
```diff
@@ -1649,6 +1660,25 @@ This function implements the following set of behaviors:
     type: Result of the conversion as described above.
 )";
 
+static const char *doc_float16_array_t = R"(
+Converts the provided Dr.Jit array/tensor type into an 16 bit floating point version.
```
wjakob, line 1664: "a 16 bit" (without "n")?

src/python/docstr.h
```diff
@@ -3939,6 +3969,200 @@ static const char *doc_PCG32_state =
...
+    :py:func:`eval_cuda()` should be used. Note that the texture is still
+    differentiable even when migrated. The :py:func:`value()` and :py:func:`tensor()`
+    operations will perform a reverse migration in this case.
+)";
+
+static const char* doc_Texture_set_value = R"(
+    Override the texture contents with the provided linearized 1D array.
+
+    When migrate is set to ``True`` on CUDA mode, the texture information
+    is fully migrated to GPU texture memory to avoid redundant storage.
```
wjakob, line 4016: It may not be clear to the user what happens with `migrate=False` -- who holds the extra copy? (They might think this refers to destroying or not destroying the function argument provided to `set_value()`, which I don't think is what this does)

docs/reference.rst
wjakob: I think the `is_half_v` operation needs to be added somewhere in this file.

docs/_static/my_theme.css
```diff
@@ -0,0 +1,3 @@
+.sidebar-container, .toc-drawer {
```
wjakob, line 1: Curious about this :)

docs/textures.rst
```diff
@@ -0,0 +1,128 @@
...
+---------------------
+
+Dr.Jit textures targeting the CUDA backend can benefit from hardware-accelerated 
+texture lookups. Internally, textures initialized with `use_accel=True` will
+create an associated *CUDA texture object* that leverages GPU hardware intrinsics 
+to perform sampling
+
+.. code-block:: python
+
+   tex = dr.cuda.Texture2f(tensor_data, use_accel=True)
```
wjakob, line 61: This is the default, right? It might be worth pointing this out.

Another note worth mentioning is that the bilinear interpolation weights used to blend texels are rounded to a lower precision when using the CUDA texture units.  12 mantissa bits, I believe? I think the details are in the CUDA C programming manual or the PTX reference.

tests/test_tensor.py
```diff
@@ -330,7 +330,7 @@ def test10_inplace(t):
     assert str(v1) == "[[3, 5],\n [9, 11]]"
     assert v3 is v1 and v3 is not v2
 
-    if dr.is_float_v(t):
+    if dr.is_float_v(t) and not dr.is_half_v(t):
```
njroussel, line 333: Shouldn't this be a check in the pytest wrapper ? Also, I'm not sure I understand why this test wouldn't work with half precision.

include/drjit/texture.h
```diff
@@ -227,37 +238,40 @@ template <typename Value, size_t Dimension> class Texture {
...
+            size_t current_shape[Dimension + 1];
 
-                bool shape_changed = false;
-                for (size_t i = 0; i < Dimension + 1; ++i) {
-                    if (current_shape[i] != tensor.shape(i)) {
-                        shape_changed = true;
-                        break;
-                    }
+            if (HasCudaTexture && is_inplace_update) {
+                jit_cuda_tex_get_shape(Dimension, m_handle, current_shape);
```
njroussel, line 247: Shouldn't this also be guarded by `m_use_accel` ?

I'm not sure I understand why and what changed in function's logic in the first place.

include/drjit/tensor.h
```diff
@@ -156,11 +156,8 @@ struct Tensor
         for (size_t i = 0; i < ndim; ++i)
             size *= shape[i];
         if (size != m_array.size()) {
-            if (m_array.size() == 1)
-                resize(m_array, size);
```
njroussel, line 160: Didn't this guarantee that we could broadcast-initialize a tensor?
Or is this handled elsewhere now?

include/drjit/python.h
```diff
@@ -748,7 +785,7 @@ nanobind::object bind_array(ArrayBinding &b, nanobind::handle scope = {},
             if constexpr (T::IsIntegral)
                 bind_int_arithmetic<T>(b);
 
-            if constexpr (T::IsFloat)
+            if constexpr (T::IsFloat && !drjit::is_half_array_v<T>)
                 bind_float_arithmetic<T>(b);
```
njroussel, line 789: Are these bound somewhere else? Where?

njroussel (approved):
Looks good to me, I only have a few questions/clarifications - I haven't seen much of these the interfaces so I feel like some stuff is missing.

tests/test_tensor.py
rtabbara: The issue was the subsequent line `dr.int_array_t` which will attempt to fetch a `int16` type which isn't bound. I've changed that though to use `dr.int32_array_t`

include/drjit/texture.h
rtabbara: Yes, thanks!

When I extended the tests to also test the LLVM variant, I noticed that `set_tensor` wouldn't work in the case where the new shape is mismatched so that's why I pulled out this logic here.

include/drjit/python.h
njroussel: Ohh I see. I thought this also included trivial operations `+`/`*`/`/`, but it's mostly trigonometric operations.

In that case, I think this is fine :+1:

wjakob: That's sounds problematic to me. It means that Dr.Jit will have to switch to its default implementation when an operation isn't defined:  it will recurse through the array and re-try running the operation on the exracted element, which would mean potentially millions of scalar PCI-Express memory transactions. Could it be fixed by updating problematic methods as follows?

```
diff --git a/include/drjit/math.h b/include/drjit/math.h
index 2dfe884f..e72399e0 100644
--- a/include/drjit/math.h
+++ b/include/drjit/math.h
@@ -596,6 +596,8 @@ template <typename Y, typename X> expr_t<Y, X> atan2(const Y &y, const X &x) {
 template <typename Value> Value atan(const Value &x) {
     if constexpr (is_detected_v<detail::has_atan, Value>)
         return x.atan_();
+    else if constexpr (is_half_v<Value>)
+        return (Value) atan((float32_array_t<Value>) x);
     else
         return atan2(x, Value(1));
 }
 ```

Comment (wjakob) [reply include/drjit/python.h:789]: About the array constants:

> So you would have to manually construct all these constants.

Let's manually construct all these constants. It should be easy to get them.

```
>>> def float_to_hex(h):
...     return hex(struct.unpack('<H', struct.pack('<e', h))[0])
...
>>> float_to_hex(np.float16(np.e))
'0x4170'
```

njroussel: Aha, I see :+1: . Indeed, we should change it then. I mistakenly thought that the templating would figure out that the best candidate is an upcast to `Float`, but that can't happen here because we need to explicitly define bindings for every type...

Issue #213, opened 2024-01-30
Loop title prevents kernel caching


merlinND:
Hi,

The following code:

```py
dr.set_log_level(dr.LogLevel.Info)

def fun(title):
    i = dr.cuda.UInt32([1, 2, 3])
    l = dr.cuda.Loop(title, (lambda: i))

    while l(i < 5):
        i += 1

    print(i)

fun("title1")
fun("title1")
fun("title1")
fun("title2")
fun("title3")
```
leads to a new kernel compilation for each new loop title:

```
jit_eval(): launching 1 kernel.
  -> launching 809d2332b518e3e1 (n=3, in=1, out=1, ops=15, jit=8.691 us):
     cache miss, build: 3.79115 ms, 3.289 KiB.
jit_eval(): done.
[5, 5, 5]
jit_eval(): launching 1 kernel.
  -> launching 809d2332b518e3e1 (n=3, in=1, out=1, ops=15, jit=8.4 us):
jit_eval(): done.
[5, 5, 5]
jit_eval(): launching 1 kernel.
  -> launching 809d2332b518e3e1 (n=3, in=1, out=1, ops=15, jit=6.61 us):
jit_eval(): done.
[5, 5, 5]
jit_eval(): launching 1 kernel.
  -> launching d083e43abc706a13 (n=3, in=1, out=1, ops=15, jit=5.29 us):
     cache miss, build: 3.86396 ms, 3.289 KiB.
jit_eval(): done.
[5, 5, 5]
jit_eval(): launching 1 kernel.
  -> launching ff86fbf889595361 (n=3, in=1, out=1, ops=15, jit=5.97 us):
     cache miss, build: 3.9326 ms, 3.289 KiB.
jit_eval(): done.
[5, 5, 5]
```

My understanding is that this is due to the loop title appearing all the way down in the JITted code. To avoid this, we could:
1. Change the emitted code to not include the loop title
2. Or, at least, change the internal usage of loops to have a constant loop title,  for example in `dr::binary_search`:

https://github.com/mitsuba-renderer/drjit/blob/8f0976008f3662756bb078f713e383a98f944e1d/include/drjit/util.h#L187-L193

What do you think?

njroussel:
Hi @merlinND 

[...]

PR #214, opened 2024-01-30
binary_search: fix unnecessary recompilations

merlinND:
This PR avoid baking the `start_` and `end_` ints into the generated binary search code in JIT mode, and switches to a constant loop title (see #213).

### DIFF
--- a/include/drjit/util.h
+++ b/include/drjit/util.h
@@ -174,7 +174,8 @@ Index binary_search(scalar_t<Index> start_, scalar_t<Index> end_,
     scalar_t<Index> iterations = (start_ < end_) ?
         (log2i(end_ - start_) + 1) : 0;
 
-    Index start(start_), end(end_);
+    Index start = opaque<Index>(start_);
+    Index end = opaque<Index>(end_);
 
     using Mask = mask_t<Index>;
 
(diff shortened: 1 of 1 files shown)


include/drjit/util.h
```diff
@@ -183,17 +184,13 @@ Index binary_search(scalar_t<Index> start_, scalar_t<Index> end_,
         using Index1 = detached_t<std::conditional_t<is_static_array_v<Index>,
                                                      value_t<Index>, Index>>;
         using Mask1 = mask_t<Index1>;
+        Index1 iterations1 = opaque<Index1>(iterations);
```
njroussel, line 187: Let's move this down into the `if` scope. We'll avoid a useless memcpy to the device if the `if` below condition isn't true.

njroussel (approved):
Tiny suggestion. 

The rest looks good to me.

include/drjit/util.h
merlinND: Good point, I've force-pushed with that change.

njroussel:
Thank you!

As mentioned in the issue, this will be refactored down the line anyway, but is still a valuable fix for the pre-nanobind versions of Mitsuba/Dr.Jit.

PR #216, opened 2024-02-22
Minor fixes to CMake build

njroussel:
Previous two commits introduced minor bugs


wjakob:
oops, all great catches -- thank you for fixing them.

PR #218, opened 2024-02-23
Fixes to bindings, and build system

njroussel:
This PR introduces the following changes:
* Fixes such that `bind_array_t<const T*>` works (const qualifiers are lost in casts from nanobind)
* Handling of already bound types in `bind_array_t`
* ~~Project-relative copy paths for Python files in the CMake build~~ replaced by 554458219323cb6e06b60aff927c43393a2eddfc

~~It also updates `drjit-core` to point to: https://github.com/mitsuba-renderer/drjit-core/pull/81~~

### DIFF
--- a/include/drjit/python.h
+++ b/include/drjit/python.h
@@ -442,8 +442,11 @@ template <typename T> NB_INLINE void bind_base(ArrayBinding &b) {
             bool success = value != Py_None && in.from_python(
                 value, (uint8_t) nb::detail::cast_flags::convert, &cleanup);
             if (success) {
-                using Out = std::conditional_t<std::is_pointer_v<Value>, Value, Value &>;
-                inst->set_entry(i, in.operator Out());
+                using Intrinsic = nb::detail::intrinsic_t<Value>;
+                using Out = std::conditional_t<std::is_pointer_v<Value>, Intrinsic*, Intrinsic &>;
+
+                Out out = in.operator Out();
+                inst->set_entry(i, (value_t<T>) out);
             }
             cleanup.release();
 
--- a/src/python/bind.cpp
+++ b/src/python/bind.cpp
@@ -185,7 +185,14 @@ nb::object bind(const ArrayBinding &b) {
 
     d.base_py = (PyTypeObject *) base_o.ptr();
 
-    // Create the type and update its supplemental information
+    // Type was already bound, let's create an alias
+    nb::handle existing = nb::detail::nb_type_lookup(b.array_type);
+    if (existing) {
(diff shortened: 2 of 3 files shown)


src/python/bind.cpp
```diff
@@ -65,7 +65,14 @@ nb::object bind(const ArrayBinding &b) {
         d.scope = meta_get_module(b).ptr();
     d.base_py = (PyTypeObject *) array_base.ptr();
 
-    // Create the type and update its supplemental information
+    // Type was already bound, let's create an alias
+    nb::handle existing = nb::detail::nb_type_lookup(b.array_type);
+    if (existing) {
+        nb::handle(d.scope).attr(name) = existing;
+        return nb::steal(existing);
```
wjakob, line 72: Are you sure that stealing the reference is the right thing to do here? `nb_type_lookup` does not increase the reference count of the returned type object. I think you need `nb:borrow` instead.

src/python/CMakeLists.txt
```diff
@@ -6,17 +6,17 @@ endif()
 # ------------------- Copy Python resources -------------------
 configure_file(
   ../../resources/config.py.in
-  ${CMAKE_BINARY_DIR}/drjit/config.py
```
wjakob, line 9: Isn't `CMAKE_BINARY_DIR` relative to the outermost `project()`? (In which case this should work even when embedded as a submodule in Mitsuba). Or am I confused?

include/drjit/python.h
```diff
@@ -442,8 +442,11 @@ template <typename T> NB_INLINE void bind_base(ArrayBinding &b) {
             bool success = value != Py_None && in.from_python(
                 value, (uint8_t) nb::detail::cast_flags::convert, &cleanup);
             if (success) {
-                using Out = std::conditional_t<std::is_pointer_v<Value>, Value, Value &>;
-                inst->set_entry(i, in.operator Out());
+                using Casted = nb::detail::intrinsic_t<Value>;
```
wjakob, line 445: Nit picking: the past tense of "cast" is also "cast", not "casted". Perhaps call it `Intrinsic`?

wjakob:
Just some small comments, the rest looks good 👍

wjakob:
Just some small comments, the rest looks good 👍

src/python/bind.cpp
njroussel: You are correct :+1: !

njroussel:
Kind of lost track of this. I removed the CMake changes, as they are handled by 554458219323cb6e06b60aff927c43393a2eddfc now which is already in `nanobind_v2`. This PR only includes the improvements to `bind_array_t`.

PR #219, opened 2024-02-29
Fixes to DLPack interfaces

njroussel:
Fixes to the DLPack interfaces:
* Scalar tensor initialization from a object implementing the DLPack protocol was broken
* Converting to Tensorflow objects through `from_dlpack` expected to directly receive the `dlcapsule` rather than the Python object

I've also added a new set of roundtrip conversion tests with `torch`, `jax`, `numpy` and `tensorflow`.


wjakob:
Perfect, LTGM 👍

PR #220, opened 2024-03-13
Various fixes & improvements

njroussel:
Various changes:
- `allclose` handles inf comparisons
- added `dr.identity`
- handle reinterpreting pointer type as `UInt32`
- handle numpy (or otherwise) scalar arithmetic with JIT array
- `binary_search` fixed
- `copysign` fixed
- fix for initializations of scalar drjit arrays from other scalar drjit arrays
- allow `array & active` on `DiffArray` types
- minor formatting

These are small changes that have been stacking up from @rtabbara  and I. Would be good to get them in before it grows into an ever larger mix of independent fixes.

### DIFF
--- a/drjit/__init__.py
+++ b/drjit/__init__.py
@@ -157,6 +157,10 @@ def allclose(
 
         cond = abs(diff) <= abs(b) * rtol_c + atol_c
 
+        # plus/minus infinity
+        if is_float_v(a):
+            cond |= all(a == b, axis=None)
+
         if equal_nan:
             cond |= isnan(a) & isnan(b)
 
@@ -577,6 +581,36 @@ def diag(arg, /):
         return result
     else:
         raise Exception('drjit.diag(): unsupported type!')
+
+
+def identity(dtype, size=1):
+    '''
+    Return the identity array of the desired type and size
+
+    This function can create identity instances of various types. In
+    particular, ``dtype`` can be:
+
+    - A Dr.Jit matrix type (like :py:class:`drjit.cuda.Matrix4f`).
+
+    - A Dr.Jit complex type (like :py:class:`drjit.cuda.Quaternion4f`).
+
+    - Any other Dr.Jit array type. In this case this function is equivalent to ``full(dtype, 1, size)``
+
+    - A scalar Python type like ``int``, ``float``, or ``bool``. The ``size``
+      parameter is ignored in this case.
+
+    Args:
+        dtype (type): Desired Dr.Jit array type, Python scalar type, or
+          :ref:`custom data structure <custom-struct>`.
+
+        size (int): Size of the desired array | matrix
+
+    Returns:
+        object: The identity array of type ``dtype`` of size ``size``
+    '''
+    result = zeros(dtype, size)
+    result += dtype(1)
+    return result
 
 
 def trace(arg, /):
--- a/drjit/__init__.py
+++ b/drjit/__init__.py
@@ -978,7 +978,7 @@ def copysign(arg0, arg1, /):
     Returns:
         float | int | drjit.ArrayBase: The values of ``arg0`` with the sign of ``arg1``
     '''
-    arg0_a = abs(a)
+    arg0_a = abs(arg0)
     return select(arg1 >= 0, arg0_a, -arg0_a)
 
 
--- a/drjit/ast.py
+++ b/drjit/ast.py
@@ -274,7 +274,7 @@ class _SyntaxVisitor(ast.NodeTransformer):
         self.op_stack.pop()
         return result
 
-    def visit_If(self, node: ast.If) -> ast.AST | tuple[ast.AST, ...]:
+    def visit_If(self, node: ast.If) -> Union[ast.AST, tuple[ast.AST, ...]]:
         (node, state_in, state_out, hints, is_scalar) = self.rewrite_and_track(node)
 
         if is_scalar:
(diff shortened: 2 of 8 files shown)


drjit/ast.py
```diff
@@ -274,7 +274,7 @@ def visit_For(self, node: ast.For) -> ast.AST:
         self.op_stack.pop()
         return result
 
-    def visit_If(self, node: ast.If) -> ast.AST | tuple[ast.AST, ...]:
+    def visit_If(self, node: ast.If) -> Union[ast.AST, tuple[ast.AST, ...]]:
```
dvicini, line 277: Purely out of curiosity: Why not use the more modern `|` notation? It was introduced back in Python 3.10, which is already a few years old at this point.

njroussel: We ship wheels from Python 3.8 upwards :/

The alternative would be to keep the 3.10 notation and use the postponed evaluation of type information so that at least it doesn't crash when importing the module.

dvicini: Fair enough. I don't really have any insight regarding the use of older Python versions, I don't know what's a good cutoff point.

drjit/__init__.py
```diff
@@ -579,6 +583,51 @@ def diag(arg, /):
         raise Exception('drjit.diag(): unsupported type!')
 
 
+def identity(dtype, size=1):
```
wjakob, line 586: Can you check if this is subsumed by `dr.ones`? It will also make identities in the new Dr.Jit.

drjit/ast.py
wjakob: @njroussel @dvicini A lot of the typing stuff is surprisingly recent.The general situation looks as follows

- Python 3.8 will often flat-out refuse to parse a Python file containing expressions that it doesn't consider part of its grammar (e.g., `tuple[*Ts]`) or which generates a runtime error (`type_1 | type_2`). Even the postponed evaluation can save you from all of these.
- Even Python 3.9 and 3.10 lack important bits.  For example, type variable tuples (the `*Ts` expansion from the previous bullet point) weren't introduced until 3.11, and we need them for a few function signatures.
- To have things be sufficiently compatible we're using Python 3.8 typing constructs in the `.py` files.
- When working in Visual Studio code, it will render the old typing constructs as new ones. So for the user, it makes no difference.
- Dr.Jit's stubgen loads both the Python source code and extension libraries to generate a unified *stub file*. These  stub files are meant for static type checkers that aren't tied to any particular Python version. So you can actually use Python 3.11+ typing constructs in the stub file for Python 3.8. This is a legit trick even mentioned in a PEP. So the stubs and replacement rules in `drjit/stubs.pat` are generally out of sync with the Python version.

drjit/ast.py
```diff
@@ -736,7 +736,7 @@ def syntax(
     recursive: bool = False,
     print_ast: bool = False,
     print_code: bool = False,
-) -> T | Callable[[T2], T2]:
+) -> Union[T, Callable[[T2], T2]]:
```
wjakob, line 739: Good catch!

include/drjit/util.h
```diff
@@ -187,8 +187,8 @@ Index binary_search(scalar_t<Index> start_, scalar_t<Index> end_,
         if (iterations >= 2 && jit_flag(JitFlag::LoopRecord)) {
             Index1 index = zeros<Index1>(width(pred(start)));
 
-            tuple(start, end, index) = drjit::while_loop(
-                tuple(start, end, index),
+            tie(start, end, index) = drjit::while_loop(
```
wjakob, line 190: let's use `drjit::tie` to avoid ADL issues.

include/drjit/util.h
```diff
@@ -187,8 +187,8 @@ Index binary_search(scalar_t<Index> start_, scalar_t<Index> end_,
...
+                make_tuple(start, end, index),
```
wjakob, line 191: `drjit::make_tuple`

src/python/meta.cpp
```diff
@@ -269,6 +269,8 @@ ArrayMeta meta_get(nb::handle h) noexcept {
     } else if (h.is_none() || nb::type_check(tp)) {
         m.type = (uint8_t) VarType::UInt32;
         m.is_class = true;
+    } else if (PyNumber_Float(h.ptr())) {
```
wjakob, line 272: This is to handle NumPy scalars?

wjakob:
Thank you! I left a few comments.

src/python/meta.cpp
rtabbara: This is to handle the case where we multiply by for instance a numpy scalar which has type `numpy.{dtype}`. In this case, I don't believe the `ndarray_check` will succeed.

drjit/__init__.py
njroussel: No, I don't believe it is. The implicit and explicit constructors will make identities, but `dr.ones()` will resolve into `full(dtype, 1)`. Using that observation I've simplified the code quite a bit. @rtabbara this should still exactly match your initial implementation.

njroussel:
@wjakob I've added one more commit: https://github.com/mitsuba-renderer/drjit/pull/220/commits/9d0557ef7bb897c2316371411dff8ae77213a38c will block any attempt to use the `ndarray` interface when constructing Dr.Jit types from other Dr.Jit types. As discussed, this is typically undesirable because it loses gradient tracking. (I ran into this pretty quickly with Mitsuba :disappointed: )

wjakob:
Perfect, that looks like the right place to make that change.

PR #233, opened 2024-07-01
Fix dr::euler_to_quat()

Speierers:
This patch fixes the `dr::euler_to_quat()` which surprisingly had never been tested in C++.


rtabbara:
Thanks Sébastien!

PR #238, opened 2024-07-12
Fixes for AD traversal inside of symbolic scope

njroussel:
This PR is a follow-up to https://github.com/mitsuba-renderer/drjit/commit/fe2526627650045b7a34f98fc75474aeda701468. That original commit only applied to symbolic loops, in this PR I've added a similar fix for calls and conditionals.

In addition, I refactored the isolation scope creation in c++ such that we can easily force it to be marked as symbolic.

The PR contains regressions tests for calls and conditionals (these could not pass prior to this fix).

### DIFF
--- a/src/extra/cond.cpp
+++ b/src/extra/cond.cpp
@@ -142,8 +142,11 @@ static void ad_cond_symbolic(JitBackend backend, const char *label,
                              bool ad) {
     bool symbolic = jit_flag(JitFlag::SymbolicScope);
 
+    /* Postponed operations captured by the isolation scope should only
+     * be executed once we've exited the symbolic scope. We therefore
+     * need to declare the AD isolation guard before the recording guard. */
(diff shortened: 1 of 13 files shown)


src/extra/cond.cpp
```diff
@@ -178,9 +181,11 @@ static void ad_cond_symbolic(JitBackend backend, const char *label,
         }
     }
 
+    JitVar true_mask = JitVar::steal(jit_var_bool(backend, true));
+    jit_new_scope(backend);
```
njroussel, line 185: Extra tiny fix that I forgot to mention:

WIthout this change the `true_mask` variable would only be assembled in the true-branch of the `if`-statement, the false-branch would just use the default value of the register. With the fix, we guarantee that `true_mask` will be assembled before the branching instruction, so the false would get a correctly initialized register for `true_mask`.

PR #239, opened 2024-07-12
`dr.detach`: Make sure we return copy of gradient disabled variables

rtabbara:
Picked up when running through the tutorials. Suppose we have

```python
opt = mi.ad.SGD(lr=0.25, params=params)
...
y = mi.Float(2.0)
opt['y'] = y  # A copy should be made here
opt['y'] += 1.0
```
Internally, the `Optimizer` `__setitem__` would perform

```python
...
self.variables[key] = dr.detach(value, True)
```

So in the case that value is  not grad-enabled. we still want a copy of the variable to be stored rather than a reference to the original value.

### DIFF
--- a/src/python/autodiff.cpp
+++ b/src/python/autodiff.cpp
@@ -111,13 +111,17 @@ static nb::object detach(nb::handle h, bool preserve_type_ = true) {
         void operator()(nb::handle h1, nb::handle h2) override {
             const ArraySupplement &s1 = supp(h1.type()),
                                   &s2 = supp(h2.type());
-            if (s1.is_diff)
+
+            if (s2.index)
                 s2.init_index((uint32_t) s1.index(inst_ptr(h1)), inst_ptr(h2));
+            else {
+                nb::object o = nb::borrow(h1);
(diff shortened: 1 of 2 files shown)


njroussel:
I'm not sure I understand the original motivation. Why is it a non-AD type? It's initialized as `mi.Float`, so it should be, unless it's non-AD variant in which case I'm not sure what to even expect.

src/python/autodiff.cpp
```diff
@@ -113,13 +113,11 @@ static nb::object detach(nb::handle h, bool preserve_type_ = true) {
                                   &s2 = supp(h2.type());
             if (s1.is_diff)
                 s2.init_index((uint32_t) s1.index(inst_ptr(h1)), inst_ptr(h2));
+            else
+                nb::inst_copy(h2, h1);
         }
     };
 
-    if ((is_drjit_array(h) && !supp(h.type()).is_diff) ||
```
njroussel, line 119: I think we should still keep the first part of this condition, or else it will produce a copy when dealing with non-AD types.

PR #242, opened 2024-07-18
Fix color.h conditional logic

Speierers:
This patch fixes the conditional logic in the color space transformations routine for static arrays.


njroussel (approved):
LGTM :bow:

PR #243, opened 2024-07-18
Fix dr::ravel and add support for dynamic arrays

Speierers:
This patch fixes the dr::ravel implementation and adds support of `ArrayXf` types.


njroussel (approved):
I think the same bug is present in `dr::unravel`, I'll fix it. Thanks for pointing it out!

PR #244, opened 2024-07-18
Minor fixes in dr::quat_to_euler

Speierers:
This patch fixes minor issues in `dr::quat_to_euler` after the refactoring


njroussel (approved):
Thanks!

PR #246, opened 2024-07-22
Fix `detach` with nested scalar Dr.Jit types and CustomOp InputOutput fix and CustomOp InputOutput fix

rtabbara:
- For `detach`, if the transform callback is called then we know it's a Dr.Jit type
- Otherwise, for other types the `transform_unknown` object is returned which by default is the borrowed reference of the passed-in value
- In the case of a scalar Dr.Jit type, we should be performing a copy because otherwise a `move` will zero-out the original variable (e.g. `dr.scalar.Array3f`)

- The other fix to modify `add_generic` (CustomOp input/output variables) revolves around a missing corresponding JIT (primal) index being provided
- This is a problem for tensors, because `apply.cpp::transform` checks whether the transformed length of the underlying transformed array matches the original to preserve the original tensor shape
- Otherwise, the new tensor is flattened
- i.e. we need the JIT index to be there in order to determine the length of the new array

### DIFF
--- a/src/python/autodiff.cpp
+++ b/src/python/autodiff.cpp
@@ -115,8 +115,7 @@ static nb::object detach(nb::handle h, bool preserve_type_ = true) {
             if (s2.index)
                 s2.init_index((uint32_t) s1.index(inst_ptr(h1)), inst_ptr(h2));
             else {
-                nb::object o = nb::borrow(h1);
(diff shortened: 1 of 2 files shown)


njroussel (approved):
LGTM, thanks for chasing these down :rocket:

PR #247, opened 2024-07-23
`meta_promote`: Handle promotion from vec4 to quaternion

rtabbara:
Motivated by usage of quaternion with `dr.select`. We should be able to support the case where the mask is a vec4 while the other args are quaternions.


njroussel (approved):
Looks good now :D

PR #250, opened 2024-08-07
Handle implicit conversion of ndarray to tensor types

rtabbara:
The implicit cast predicate for Dr.Jit array types didn't previously handle the case where the target type is a tensor and the source is a sequence (e.g. ndarray)


njroussel (approved):
Looks good to me! Nice catch :+1:

PR #255, opened 2024-08-13
Simplify LD_PRELOAD documentation for Linux ASAN debugging

WillAyd:
Instead of asking users to find where the ASAN / stdc++ libs are on their platform, you can use gcc to simplify that process

(this may also work for macOS)


wjakob:
Nice, thank you!

PR #256, opened 2024-08-14
Handle list comprehensions in `@dr.syntax`

### LINKED ISSUE #252: Issue in `dr.syntax` transformation with scalar loops and list comprehensions
I ran into an issue when using `dr.syntax` on a longer function that mixes several types of loops (some symbolic, some scalar/evaluated) and list comprehension.

I haven't fully narrowed this down, but it seems to be some combination of having a list expression and a scalar loop reuse a variable name (here, `j`).

Here is a reproducer:

```python
import drjit as dr

@dr.syntax
def f():
  n = 10
  k = 3

  print([j for j in range(k)])

  i = dr.zeros(dr.llvm.Int32, n)
  result = dr.zeros(dr.llvm.Float)
  while i < 10:
    for j in range(k):
      result += i * j
    i += 1
  print(result)

f()
```

This prints:

```
Traceback (most recent call last):
  File ".../mitsuba3/build/drjit_loop.py", line 19, in <module>
    f()
  File ".../mitsuba3/build/drjit_loop.py", line 13, in f
    while i < 10:
UnboundLocalError: cannot access local variable 'j' where it is not associated with a value
```

njroussel:
This PR fixes #252.

List comprehensions and other generators define statement-local variables that can never be part of the state loop. This PR guarantees this by ignoring those variables names during the AST traversal of `@dr.syntax`.

I've added a test that isn't much more than a regressions test - it doesn't actually verify that the generator's variables become part of the loop state.

I think this is a fairly safe change @wjakob, but I'm also a bit adverse to introducing more and more handling of these edge cases in `@dr.syntax`

### DIFF
--- a/drjit/ast.py
+++ b/drjit/ast.py
@@ -147,6 +147,34 @@ class _SyntaxVisitor(ast.NodeTransformer):
             self.var_w.add(node.id)
         return node
 
+    def visit_comp(self,
+                   node: Union[ast.ListComp, ast.SetComp, ast.DictComp, ast.GeneratorExp]) -> ast.AST:
+        var_r, var_w = self.var_r, set(self.var_w)
+        self.var_r = set()
+        result = self.generic_visit(node)
+
+        comp_targets = set()
+        for comp in node.generators:
+            comp_targets.add(comp.target.id)
+
+        # Targets should not be considered, and no assigments can be made
+        self.var_r = (self.var_r - comp_targets) | var_r
+        self.var_w = var_w
+
+        return result
+
+    def visit_ListComp(self, node: ast.ListComp) -> ast.AST:
+        return self.visit_comp(node)
+
+    def visit_SetComp(self, node: ast.DictComp) -> ast.AST:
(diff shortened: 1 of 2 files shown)


wjakob:
Hi @njroussel,

thank you for this fix. I had not considered comprehensions when designing the AST transformations. It's a natural Python construct that should most definitely be supported.

But I'm thinking that the implementation could potentially be even easier. 

First, `comp.target.ctx` can AFAIK only be an `ast.Store`, so I'm thinking that the two cases aren't needed.

Second, you cannot make assignments to outside variables in a comprehension. So at the end, you can just set `self.var_w = var_w`. In fact, maybe nothing at all needs to be done with `self.var_w` (no copies, no re-assignment) since we know that it won't be changed.

Last, it is important when a comprehension reads additional variables, but we can exclude variables that the comprehension sets itself in its isolated scope.

How about `self.var_r = (self.var_r - comp_targets_w) | var_r`

njroussel:
Indeed that simplified the code a bit, thanks.

PR #259, opened 2024-08-16
Add `Array6*` bindings. 

rtabbara:
For Mitsuba suite of SGGX distribution functons we need this type to be exposed

### DIFF
--- a/include/drjit/array_mask.h
+++ b/include/drjit/array_mask.h
@@ -37,6 +37,30 @@ private:
     size_t index;
 };
 
+template <typename Array1, typename Array2>
+struct MaskBitPacketRecursive {
+
+    MaskBitPacketRecursive(Array1& a1, Array2& a2, size_t index) :
+        a1(a1), a2(a2), index(index) {}
+
+    operator bool() const {
+        return (index < Array1::Size) ?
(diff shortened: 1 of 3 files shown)


wjakob:
A quick question: Could we implement some kind of auto-conversion to ArrayXf or some other workaround (e.g. a tuple, or 2x Array3fs)? I don't see us using the Array6* types for actual arithmetic much, and there is a non-negligible cost to adding another one of these (plus, it's weird to have 0-4 and then 6 without 5 ;)). AFAIK we're just using this type for temporary storage/access and perhaps a tiny bit of SGD math.

wjakob:
If we absolutely need it, then the binding could be done on demand from Mitsuba, rather than Dr.Jit itself. The `drjit/include/python.h` file exposes functionality to create extra bindings when need it.

dvicini:
I think we have some flexibility in the API for the SGGX sampling functions. I wouldn't mind if those were changed to pass 2 x Array3fs. There is a natural split here in the first 3D vector representing diagonal, and the second representing off-diagonal elements

rtabbara:
Base on the discussion, I've opened up a PR in Mitsuba [1276](https://github.com/mitsuba-renderer/mitsuba3/pull/1276/) instead and will close off this PR

PR #262, opened 2024-08-20
Swizzle and implicit binding predicate fixes

rtabbara:
`dr.cross` : Cast return type back to original type
- Use of swizzle operator will return a dr.jit array type. This means for example swizzling a `mi.Vector3f` returns a `dr.*.Array3f`

For implicit conversion predicate from dr.jit types, also check dimensions are compatible
- Previously types such as `Array4f` -> `Matrix4<Color3f>` would return true as we're just simply checking that the first dimension aligns

### DIFF
--- a/src/python/base.cpp
+++ b/src/python/base.cpp
@@ -321,7 +321,9 @@ static NB_NOINLINE nb::handle tp_getattro_fallback(nb::handle h0, nb::handle h1)
                                                            : DRJIT_DYNAMIC);
 
                 try {
-                    nb::handle tp2 = meta_get_type(m2);
+                    nb::handle tp2 = swizzle_size == s0.shape[0] ?
+                        h0.type() :
+                        meta_get_type(m2);
                     const ArraySupplement &s2 = supp(tp2);
 
                     ArraySupplement::Item item = s0.item;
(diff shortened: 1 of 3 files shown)


rtabbara:
Also addresses Mitsuba issue [1280](https://github.com/mitsuba-renderer/mitsuba3/issues/1280)

wjakob:
Wouldn't it be preferable to fix this  by changing swizzling so that it preserves the type when the number of array dimensions remains the same?

Speierers:
Somehow this patch prevents casting a `Matrix<4, Float>` into a `Matrix<4, Color<3, Float>>` as in the following example:


'''
import mitsuba as mi
mi.set_variant('llvm_ad_rgb_polarized')

print(mi.Spectrum(mi.mueller.depolarizer(1)))
'''

rtabbara:
@wjakob I've updated the PR so that instead the swizzle operation returns the original type if the swizzle size is compatible

@Speierers I think I've fixed this now and added some corresponding tests related to this

njroussel:
There will always be a tradeoff between an "excessive" check in the implicit cast predicate, and letting it fail during the actual construction attempt. These change are fair for now, in my opinion.

Issue #264, opened 2024-08-22
Mismatch in float precision for (some) math functions


dvicini:
A few of the math operations in `__init__.py` unintentionally (I assume) modify the precision of the computation. This is because they retrieve some of the output types using code such as `Array3f = getattr(m, 'Array3f', None)`

Concretely, this makes it that for example `dr.quat_to_euler` returns a single precision Array3f, even when invoked on a double precision quaternion:

```
>>> type(dr.quat_to_euler(dr.llvm.Quaternion4f64(1.0)))
<class 'drjit.llvm.Array3f'>
```

This pattern appears to happen in `matrix_to_quat`, `quat_to_euler`, `euler_to_quat`, `transform_decompose` and `transform_compose` (might not be an exhaustive list).

rtabbara:
Fixed in 6e44e23

PR #265, opened 2024-08-22
Fix type fetching of some Dr.Jit Python functions

rtabbara:
Need to ensure underlying floating precision is consistent with provided arguments

Fix for #264


njroussel (approved):
All good!

PR #268, opened 2024-08-26
Fix forward differentiation of packet scatters

njroussel:
The target of a packet `scatter` operation has a special code path in the AD layer when forward differentiated (`PacketScatter::forward()`): 
* We first accumulate any existing gradients (prior to the `scatter`) on the target
* Secondly, we individually accumulate each of the packet's values' gradients

The first step was using the wrong JIT index to identify the existing gradients on the target.

(The backward traversal looked up the correct JIT index)

I added a simple regressions test.

### DIFF
--- a/src/extra/autodiff.cpp
+++ b/src/extra/autodiff.cpp
@@ -3096,8 +3096,8 @@ public:
 
         Variable *target = state[m_output_indices[0]];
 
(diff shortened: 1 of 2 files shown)


rtabbara (approved):
LGTM!

PR #269, opened 2024-08-29
Fix `dr.reinterpret_array` to vectorized class pointers

njroussel:
This PR fixes the following snippet: 
```python
uint32_values = UInt32(...)
ptr = dr.reinterpret_array(BasePtr, uint32_values)
assert isinstance(ptr, BasePtr)
```
Previously, `reinterpret_array` would return a value of type `UInt32` instead of `BasePtr`.

For this to work we must allow reinterpretations of `T::IsClass` types to (and only to) vectorized `uint32_t` types. In addition, we also cannot rely on `meta_get_type` to figure out the correct return type as it is unaware of any class types that were bound by other libraries (e.g Mitsuba).

I've added a unit test covering this behavior.

### DIFF
--- a/include/drjit/python.h
+++ b/include/drjit/python.h
@@ -561,31 +561,45 @@ template <typename T> void bind_cast(ArrayBinding &b) {
     using Float32 = float32_array_t<T>;
     using Float64 = float64_array_t<T>;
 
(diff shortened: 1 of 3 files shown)


wjakob:
This change looks good to me. Can we also cast the other way? (`UInt32` -> `BasePtr`)

njroussel:
Yes, the test covers both :+1:

wjakob:
Great, thank you!

PR #271, opened 2024-08-30
Fix `gather_` definition for packets (avx2)

DoeringChristian:
Fixes the definition of the `gather_` function in the `packet_sse4.h` file for `X86_AVX2`.


njroussel (approved):
Thank you! 

I'm not that surprised that I missed one x)

PR #280, opened 2024-09-13
Fixed reduce_identity creating intermediate opaque

DoeringChristian:
The reduce_identity function, is constructing an intermediate variable to initialize the identity element.
With JIT variants, this was constructing an intermediate opaque value via the `init_data` function.
This PR changes the type of the intermediate value from a JIT type to a scalar type, removing the need for the intermediate opaque.

Previously:
```python
In [1]: import drjit as dr

In [2]: dr.set_log_level(dr.LogLevel.Debug)

In [3]: i = dr.detail.reduce_identity(dr.cuda.Float, dr.ReduceOp.Add, 1)
jit_var_new(): float32 r1 = data(<0x302000000>)
jit_var_mem_copy(): float32 r1[1] = copy_from(host, <0x7ffccba48230>)
jit_var_new(): float32 r2 = 0
```

Now:
```python
In [1]: import drjit as dr

In [2]: dr.set_log_level(dr.LogLevel.Debug)

In [3]: i = dr.detail.reduce_identity(dr.cuda.Float, dr.ReduceOp.Add, 1)
jit_var_new(): float32 r1 = 0
```

### DIFF
--- a/src/python/detail.cpp
+++ b/src/python/detail.cpp
@@ -62,7 +62,6 @@ nb::object reduce_identity(nb::type_object_t<dr::ArrayBase> tp, ReduceOp op, uin
     ArrayMeta m { };
     m.backend = (uint64_t)JitBackend::None;
     m.ndim = 1;
-    m.is_diff = s.is_diff;
     m.type = s.type;
     m.shape[0] = DRJIT_DYNAMIC;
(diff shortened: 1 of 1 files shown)


wjakob:
Thanks!

PR #283, opened 2024-09-18
Support for indexing into zero-dim tensor

rtabbara:
A common pattern in the Mitsuba tutorials is to perform a reduction on a tensor across all axes and then fetch the single scalar value. e.g.

```python
# TensorXf type of shape=()
y = dr.mean(tensor, axis=None)

# Scalar type
x = y[0]
```

The PR adds this feature to zero-index into these tensors. Note, that for consistency compatible slices (e.g. `y[-1]`) are also still valid and slice assignment is additionally supported.


rtabbara:
After further discussions, we've decided this isn't a necessary feature

PR #287, opened 2024-09-25
CallSupport: allow calls on derived classes

merlinND:
The DrJit registry remains untouched, we simply allow instantiating a `CallSupport<Child>` which exposes the interface of  both `Child` and its parent class.

### DIFF
--- a/docs/cpp.rst
+++ b/docs/cpp.rst
@@ -172,7 +172,7 @@ Python interface to C++.
 The argument ``args`` must always be a tuple that will be unpacked and passed
 as arguments of ``true_fn`` and ``false_fn``. The return value of these
 function can be any tree of arbitrarily nested arrays, tuples, and other
-:ref:`custom data structures <custom_types_cpp>`. 
+:ref:`custom data structures <custom_types_cpp>`.
 
 The short-hand notation provided through the :py:func:`@drjit.syntax
 <drjit.syntax>` decorator is not available in C++.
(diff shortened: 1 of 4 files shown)


docs/cpp.rst
```diff
@@ -322,6 +322,20 @@ on a Dr.Jit instance arrays. Below is an overview of the available macros:
...
    Demarcates the end of an interface block.
 
+.. c:macro:: DRJIT_CALL_TEMPLATE_INHERITED_BEGIN(Name, Parent)
+
+   A variant of the above macro that should be used when ``Name`` refers to a
+   template class that inherits from another template class ``Parent``.
+   The class ``Name`` should only be registered under the registry domain of ``Parent``.
+   The interface of ``Parent`` must have already been exposed with a call to
+   ``DRJIT_CALL_TEMPLATE_BEGIN(Parent) ... DRJIT_CALL_END()``.
+   The interface will then be extended with any methods listed in this block.
```
wjakob, line 332: Could you add that it's only safe to use this feature in polymorphic types, where C++ can use RTTI to determine if a particular type is a subclass of `Parent` or `Name`?

wjakob (commented):
Could I ask you add tests in the Dr.Jit repository? There is a file `tests/call_ext.cpp` and associated Python script that tests the functionality in a self-contained way here without depending on Mitsuba.

merlinND:
Added this feature to the test bindings. I'll squash the commit history before merging once CI is green, and let me know if you'd like any other change!

wjakob:
Excellent, thank you!

PR #288, opened 2024-09-25
TraversableBase intrusive interface

DoeringChristian:
This PR adds a `TraversableBase` class with the `traverse_1_cb_ro` and `traverse_1_cb_rw` functions and implements the `nanobind::intrusive_base` class.
The class can be implemented for any sub-class that should be traversable, and the functions can be generated by the `DR_TRAVERSE` macros.
It also changes the `TraverseCallback::operator()` function, which now returns a `uint64_t`. This is required to implement a `make_opaque` function, that is able to traverse `C++` objects.

### DIFF
--- a/include/drjit/traversable_base.h
+++ b/include/drjit/traversable_base.h
@@ -1,0 +1,79 @@
+#pragma once
+
+#include "array_traverse.h"
+#include "drjit-core/macros.h"
+#include "nanobind/intrusive/counter.h"
+#include "nanobind/intrusive/ref.h"
+#include <drjit-core/jit.h>
+#include <drjit/map.h>
+
+NAMESPACE_BEGIN(drjit)
+
+/// Interface for traversing C++ objects.
+struct TraversableBase : nanobind::intrusive_base {
+    virtual void traverse_1_cb_ro(void *, void (*)(void *, uint64_t)) const = 0;
+    virtual void traverse_1_cb_rw(void *, uint64_t (*)(void *, uint64_t))   = 0;
+};
+
+/// Macro for generating call to traverse_1_fn_ro for a class member
+#define DR_TRAVERSE_MEMBER_RO(member)                                          \
+    drjit::log_member_open(false, #member);                                    \
+    drjit::traverse_1_fn_ro(member, payload, fn);                              \
+    drjit::log_member_close();
+/// Macro for generating call to traverse_1_fn_rw for a class member
+#define DR_TRAVERSE_MEMBER_RW(member)                                          \
+    drjit::log_member_open(true, #member);                                     \
+    drjit::traverse_1_fn_rw(member, payload, fn);                              \
+    drjit::log_member_close();
+
+inline void log_member_open(bool rw, const char *member) {
+    jit_log(LogLevel::Debug, "%s%s{", rw ? "rw " : "ro ", member);
+}
+
+inline void log_member_close() { jit_log(LogLevel::Debug, "}"); }
+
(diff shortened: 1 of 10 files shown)


DoeringChristian:
@merlinND, @wjakob This PR should be ready for review.
Note, there is one change from what we discussed last Friday.
I had to add a boolean to the argument of the `traverse` function, indicating  whether `traverse_cb_ro` or `traverse_cb_rw` should be called. We cannot use the output of the callback to determine if the variable index should be overwritten, since overwriting it with $0$ might be desired (there is a test that fails in that case).

include/drjit/traversable_base.h
```diff
@@ -0,0 +1,79 @@
...
+        if constexpr (!std ::is_same_v<Base, drjit ::TraversableBase>)         \
+            Base ::traverse_1_cb_rw(payload, fn);                              \
+        drjit::traverse_py_cb_rw(this, payload, fn);                           \
+    }
+
+#if defined(_MSC_VER)
+#define DRJIT_EXPORT __declspec(dllexport)
+#else
+#define DRJIT_EXPORT __attribute__((visibility("default")))
+#endif
```
merlinND, line 77: Isn't this already defined somewhere else?

src/python/apply.h
```diff
@@ -93,8 +93,8 @@ struct TransformPairCallback {
 };
 
 /// Invoke the given callback on leaf elements of the pytree 'h'
-extern void traverse(const char *op, TraverseCallback &callback,
-                     nb::handle h);
+extern void traverse(const char *op, TraverseCallback &callback, nb::handle h,
+                     bool rw = false);
```
merlinND, line 97: Document the `rw` parameter, explain why it's needed and what the user should pass.

src/python/apply.h
```diff
@@ -57,7 +57,7 @@ struct TraverseCallback {
     // Type-erased form which is needed in some cases to traverse into opaque
     // C++ code. This one just gets called with Jit/AD variable indices, an
     // associated Python/ instance/type is not available.
-    virtual void operator()(uint64_t index);
+    virtual uint64_t operator()(uint64_t index);
```
merlinND, line 60: Document the return value.

src/python/apply.cpp
```diff
@@ -614,30 +615,32 @@ void traverse(const char *op, TraverseCallback &tc, nb::handle h) {
...
                 for (nb::handle field : df) {
                     nb::object k = field.attr(DR_STR(name));
-                    traverse(op, tc, nb::getattr(h, k));
+                    traverse(op, tc, nb::getattr(h, k), rw);
                 }
-            } else if (nb::object cb = get_traverse_cb_ro(tp); cb.is_valid()) {
+            } else if (nb::object cb = get_traverse_cb_ro(tp); cb.is_valid() && !rw) {
                 cb(h, nb::cpp_function([&](uint64_t index) { tc(index); }));
+            } else if (nb::object cb = get_traverse_cb_rw(tp); cb.is_valid() && rw) {
+                cb(h, nb::cpp_function([&](uint64_t index) { return tc(index); }));
```
merlinND, line 643: If the user passed `rw = true` but only `get_traverse_cb_ro(tp)` is valid, not `get_traverse_cb_rw(tp)`, is that an error? Should we throw?

merlinND (commented):
Hi @DoeringChristian,

Thanks for this concise PR! The code looks good to me overall, but I'm not so familiar with this part of the codebase so @wjakob will probably need to take a look as well.

njroussel:
Closing, superseded by #336

PR #289, opened 2024-09-25
Changing default behaviour of reductions for tensor types.

rtabbara:
Changing default behaviour of reductions for tensor types.

To be consistent with past releases, we make `axis=None` for tensor types, and `axis=0` otherwise. A default value of `axis=...` is used to capture this behaviour.

Also move `mean` to cpp implementation

### DIFF
--- a/tests/test_reduction.py
+++ b/tests/test_reduction.py
@@ -96,11 +96,11 @@ def test04_sum(t):
     m = sys.modules[t.__module__]
     assert dr.allclose(dr.sum(6.0), 6)
 
-    a = dr.sum(m.Float([1, 2, 3]), mode='evaluated')
+    a = dr.sum(m.Float([1, 2, 3]), axis=0, mode='evaluated')
     assert dr.allclose(a, 6)
     assert type(a) is m.Float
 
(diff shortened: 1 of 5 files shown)


tests/test_reduction.py
```diff
@@ -96,11 +96,11 @@ def test04_sum(t):
     m = sys.modules[t.__module__]
     assert dr.allclose(dr.sum(6.0), 6)
 
-    a = dr.sum(m.Float([1, 2, 3]), mode='evaluated')
+    a = dr.sum(m.Float([1, 2, 3]), axis=0, mode='evaluated')
```
njroussel, line 99: This one and the one below should be optional right? The default should be zero here?

njroussel (approved):
LGTM, thank you :100: 

One tiny clarification - I think I'm missing something.

tests/test_reduction.py
rtabbara: Ah yep, you're right. This was a remnant from the first iteration of the change where I had split the functions up. I've removed this now because it's not needed.

wjakob:
Hi @rtabbara -- could you take a look at `changelog.rst` as well? It covers changes related to reductions and is probably not accurate anymore with this PR merged.

PR #293, opened 2024-09-27
PyPI wheels

njroussel:
This PR introduces changes to successfully build wheels for PyPI again.

It does the following:
* Transfer scattered package metadata to a single source (`pyproject.toml`)
* Support for native MacOS arm65 builds (no more cross-compilation)
* Fixes minor issues in source code that lead to unstable wheels
* Disables CPython stable ABI builds, we have separate wheels for Python version 3.8 to 3.13 (included)

These changes were tested over on https://github.com/njroussel/drjit-fork, with packages uploaded to https://test.pypi.org/project/throwaway8213fork/ (`pip install --index-url https://test.pypi.org/simple/ throwaway8213fork`)


njroussel:
Pushed one more commit that additionally publishes an `*-abi3` wheel (stable ABI release).

njroussel:
Added a small document that will serve as a checklist when publishing new releases.

njroussel:
And another commit to add `dr.get_cmake_dir()` as we did previously.

src/python/base.cpp
```diff
@@ -992,20 +992,33 @@ static VarState get_state(nb::handle h_) {
 }
 
 /// Create interned string for a few very commonly used identifiers
```
rtabbara, line 994: Not super necessary but maybe small inline comment about why initialization was moved from static because on Windows GIL may not have been acquired at this point.

rtabbara (approved):
LGTM! :shipit: Thanks for writing up the release doc too :bow:

src/python/base.cpp
njroussel: You're right, done :+1:

PR #297, opened 2024-10-10
Fix scatter for JIT arrays

YunHsiao:
For JITArray `set_entry` is needed to write values.

### DIFF
--- a/include/drjit/array_router.h
+++ b/include/drjit/array_router.h
@@ -1065,7 +1065,7 @@ void scatter(Target &target, const Value &value, const Index &index,
 
         if (mask) {
             if constexpr (is_array_v<Target>)
-                target[index] = value;
+                target.set_entry(index, value);
             else
                 ((Value *) target)[index] = value;
         }


njroussel:
Hi @YunHsiao 

Could you share a small C++ snippet that currently fails without your suggested change? I'm failing to see what is currently missing.

YunHsiao:
Sure, it could boil down to something like this:

```cpp
template<typename Value>
struct MyPoint2D {
    using Float = drjit::float_array_t<Value>;
    Float x;
    Float y;

    DRJIT_STRUCT(MyPoint2D, x, y);
};

int main()
{
    // Init following data
    MyPoint2D<CUDAArray<float>> list;
    MyPoint2D<float> slice;

    // Compile-time error, can't write to r-value
    drjit::scatter(list, slice, 0);

    return 0;
}
```

njroussel:
Aha I see, I had never considered that possibility - I tend to always convert the values to JIT types before scattering. Thanks !

njroussel (approved):

Issue #253, opened 2024-08-12
dr.syntax: AD gets disabled by variable use in while loop


dvicini:
I have some code that mixes AD with handwritten derivatives & loops (e.g., similar to something like PRB)

It appears that within `dr.syntax`, any use of a differentiable variable within a loop disables the variable's AD graph.

Here is an example:
```python
import drjit as dr

@dr.syntax
def f():

  i = dr.zeros(dr.llvm.Int32, 10)
  result = dr.zeros(dr.llvm.ad.Float, 10)

  a = dr.linspace(dr.llvm.ad.Float, 0, 1, 10)
  dr.enable_grad(a)
  print(dr.grad_enabled(a))

  with dr.suspend_grad():
    while i < 5:
      result += a

  print(dr.grad_enabled(a))

f()
```

This prints
```
True
False
```
But I would have expected this to print
```
True
True
```
The current behavior is a bit unintuitive, and leads to confusing loss of gradient tracking. To me it seems that the loop should have no influence on whether `a` has gradients enabled or not, similar to the pre-dr.syntax behavior.

njroussel:
Hi @dvicini 

It looks like the syntax rewriting is a bit aggressive here: it thinks `a` is part of the loop state. This means that `a` gets re-assigned at the end the loop and hence within the `suspend_grad`. Chaning the loop to ` while dr.hint(i < 5, exclude=[a]): ` fixes the issue.

I'll look into why it thinks `a` should be in the `state`. As a general rule, the `@dr.syntax` tends to be "safer" than necessary in order to guarantee the loop is valid, but that can come at a cost as you see here.

njroussel:
Ok, there's not much we can do here, I believe. 

In short, even though it's clear in your reproducer that `a` is not being written to in the loop, we cannot guarantee that in a more general case. For example, a random number generator will effectively never be on the left-hand side of an assignement  (i.e `sampler =  sampler.next_1d()`), but it must still be considered as part of the loop state in order to work because it might evolve implicitly (i.e `sampler.next_1d()`). So, in this case, we still consider `a` to be in the loop state because even though it's only use on the right-hand side of an assignment, it might have evolved implicitly.

The workflow for these kind of situations is usually to add `print_code=True` to `@dr.syntax()` and look at the rewritten function. It's usually fairly obvious if too many variables are included and that you should then specify then in the exclude list of a `dr.hint` statement.

dvicini:
[...]

PR #299, opened 2024-10-16
Improve the behavior of control flow operations (``dr.if_stmt(), drjit.while_loop()``) in AD-suspended scopes

wjakob:
Dr.Jit control flow operations (``dr.if_stmt(), drjit.while_loop()``) currently disable gradient tracking of all variable state when the operation takes place within an AD-disabled scope.

This can be surprising when a ``@dr.syntax`` transformation silently passes local variables to such an operation, which then become non-differentiable. @dvicini reported this in issue https://github.com/mitsuba-renderer/drjit/issues/253.

This commit carves out an exception: when variables aren't actually modified by the control flow operation, they can retain their AD identity.

The PR has 3 parts:

- the first commit changes the semantics of ``ad_var_inc_ref`` so that it only does reference counting. The old behavior (drop ref if variable is currently non-differentiable) is moved to a new function ``ad_var_copy_ref()`` (see the description of this commit for details). This already fixes 80% of the problems.
- Parts 2 and 3 fix ``dr.if_stmt()`` and ``dr.while_loop()``, respectively.

The PR depends on a Dr.Jit-Core PR: https://github.com/mitsuba-renderer/drjit-core/pull/104

### DIFF
--- a/include/drjit/autodiff.h
+++ b/include/drjit/autodiff.h
@@ -123,12 +123,10 @@ struct DRJIT_TRIVIAL_ABI DiffArray
     }
 
     DiffArray(const DiffArray &a) {
-        if constexpr (IsFloat) {
-            m_index = ad_var_inc_ref(a.m_index);
-        } else {
-            m_index = a.m_index;
-            jit_var_inc_ref(m_index);
-        }
+        if constexpr (IsFloat)
+            m_index = ad_var_copy_ref(a.m_index);
+        else
+            m_index = jit_var_inc_ref(a.m_index);
     }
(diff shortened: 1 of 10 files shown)


njroussel (approved):
Looks good to me. It was not immediately obvious to me what would happen with a `dr.scatter()` that only involved attached state variables - but it's actually fine.

andyyankai:
Hi, may I ask if this fix can relate to this issue? Thanks!
https://github.com/mitsuba-renderer/mitsuba3/issues/1334

PR #302, opened 2024-10-18
fix mask type select

YunHsiao:
`select` doesn't quite work when source types are masks:

```cpp
	drjit::Packet<float, 16> a, b;
	auto m = drjit::select(a > 0, a < b, a > b); // Return type is deduced back to Packet instead of PacketMask
```

And for some compilers this is a straight-up compile error:

```
no matching conversion for static_cast from 'const drjit::PacketMask<float, 16>' to 'ref_cast_t<drjit::PacketMask<float, 16>, E>' (aka 'drji
t::Packet<float, 16>')
```

### DIFF
--- a/include/drjit/array_router.h
+++ b/include/drjit/array_router.h
@@ -264,8 +264,9 @@ DRJIT_INLINE auto select(const M &m, const T &t, const F &f) {
             });
         return result;
     } else {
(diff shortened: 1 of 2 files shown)


njroussel (approved):
Thanks :+1:, looks good to me.

PR #304, opened 2024-10-22
Support higher axis numbers for dr.any and dr.all reductions

zhestyatsky:
Hello,

I have been experimenting with the latest drjit version and noticed that `dr.all()` and `dr.any()` operations seem to not be supported for Tensors when they are evaluated with `axis` argument greater than `0`.

Reproducer below raises `RuntimeError: drjit.all(<drjit.cuda.TensorXb>): tensor type is not compatible with the requested reduction.

```py
import drjit as dr
from drjit.cuda import TensorXb

v = TensorXb([[True, False, True],
              [True, True, True],
              [True, False, True],
              [True, True, True]
             ])
dr.all(v, axis=1)
```

Similar exception is raised when calling `dr.any()`.

This PR aims to enable `dr.all()` and `dr.any()` operations for Tensors with `axis` parameter greater than `0`.

### DIFF
--- a/src/python/reduce.cpp
+++ b/src/python/reduce.cpp
@@ -311,15 +311,23 @@ nb::object reduce(uint32_t op, nb::handle h, nb::handle axis_, nb::handle mode) 
                 value = reduce(op, value, axis, mode);
                 return tp(value, nb::tuple());
             } else {
-                if (op >= (uint32_t) ReduceOp::Count) {
+                if (op == (uint32_t) ReduceOpExt::Count) {
                     if (axis_len == 1 && red_axis == 0)
                         return reduce_seq(op, h, nb::int_(0), mode);
                     nb::raise_type_error("tensor type is not compatible with "
                                          "the requested reduction.");
                 }
                 // Complex case, defer to a separate Python implementation
+                ReduceOp reduceOp;
+                if (op < (uint32_t) ReduceOp::Count) {
+                    reduceOp = static_cast<ReduceOp>(op);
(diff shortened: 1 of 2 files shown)


src/python/reduce.cpp
```diff
@@ -311,15 +311,23 @@ nb::object reduce(uint32_t op, nb::handle h, nb::handle axis_, nb::handle mode)
...
             } else {
-                if (op >= (uint32_t) ReduceOp::Count) {
+                if (op == (uint32_t) ReduceOpExt::Count) {
                     if (axis_len == 1 && red_axis == 0)
                         return reduce_seq(op, h, nb::int_(0), mode);
                     nb::raise_type_error("tensor type is not compatible with "
                                          "the requested reduction.");
                 }
                 // Complex case, defer to a separate Python implementation
+                ReduceOp reduceOp;
```
wjakob, line 321: Very minor. `rendceOp` -> `reduce_op` (underscore case for vriables)

wjakob:
Hi @zhestyatsky -- is this PR ready to be merged or still under development? It states "Draft" status.
One thing I am surprised about is that this works as-is.

The all/and reductions are AFAIK only implemented for 32-bit integers. We previously got away with using them for reduction down to 1 element by applying a simple follow-up 4->1 element reduction and doing something some extra memsets in case the input array wasn't exactly a multiple of 4 elements.

For reductions along another axes, I am wondering if there won't be mix-ups when the element count isn't a multiple of 4?

Issue #303, opened 2024-10-19
drjit.gather does not support PyTrees defined with @dataclass


zihay:
Hi,

I'm using Dr.Jit for my project and using PyTrees as described in the documentation. I expect that drjit.gather should work with data classes defined using the `@dataclass` decorator. However, I'm encountering an issue where it does not. 
**Code that doesn't work:**
```python
from dataclasses import dataclass
import drjit as dr
from drjit.cuda.ad import Array2, Int

def test_bbox():
    @dataclass
    class BoundingBox:
        p_min: Array2
        p_max: Array2

    boxes = dr.zeros(BoundingBox, 8)
    box = dr.gather(BoundingBox, boxes, Int(1))
    print(box.p_min)
    print(box.p_max)

test_bbox()
```
**Error message:**
```
TypeError: drjit.gather(<__main__.BoundingBox>): unsupported dtype!
```
**Code that works**
```python
import drjit as dr
from drjit.cuda.ad import Array2, Int

def test_bbox():
    class BoundingBox:
        DRJIT_STRUCT = {'p_min': Array2, 'p_max': Array2}
        p_min: Array2
        p_max: Array2

    boxes = dr.zeros(BoundingBox, 8)
    box = dr.gather(BoundingBox, boxes, Int(1))
    print(box.p_min)
    print(box.p_max)

test_bbox()
```
In this version, I define the `BoundingBox` class with the `DRJIT_STRUCT` annotation, and `drjit.gather` works as expected.
Based on the documentation, I expect that data classes should be fully supported in operations like `drjit.gather`, similar to custom classes with `DRJIT_STRUCT`. Is this a bug, or is there an additional step required to make data classes compatible with drjit.gather?
Thank you for your assistance!

rtabbara:
Hi @zihay ,

Thanks for reporting this. This feature has now been added into master (PR #305)

PR #305, opened 2024-10-22
Add support for gather/scatter of dataclass

rtabbara:
Motivated by #303 , add support for `gather` and `scatter` when using `dataclass` types

### DIFF
--- a/src/python/memop.cpp
+++ b/src/python/memop.cpp
@@ -61,18 +61,24 @@ nb::object gather(nb::type_object dtype, nb::object source,
 
             return std::move(result);
         } else {
-            nb::object dstruct = nb::getattr(dtype, "DRJIT_STRUCT", nb::handle());
-            if (dstruct.is_valid() && dstruct.type().is(&PyDict_Type)) {
-                nb::dict dstruct_dict = nb::borrow<nb::dict>(dstruct);
+            if (nb::dict ds = get_drjit_struct(dtype); ds.is_valid()) {
                 nb::object out = dtype();
-
-                for (auto [k, v] : dstruct_dict) {
+                for (auto [k, v] : ds) {
                     if (!v.is_type())
                         throw nb::type_error("DRJIT_STRUCT invalid, expected types!");
                     nb::type_object sub_dtype = nb::borrow<nb::type_object>(v);
                     nb::setattr(out, k, gather(sub_dtype, nb::getattr(source, k), index, active, mode));
                 }
                 return out;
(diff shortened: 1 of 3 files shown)


merlinND:
Very cool!
Could you please check if this also works with `__future__`-style annotations?
In the unit tests, that might require a separate `test_*.py` file since `from __future__ import annotations` affects the whole file I think.

rtabbara:
Hey @merlinND 

I've tested it out locally and it works but with caveats. There's a more general issue that was uncovered when using `from __future__ import annotations` to lazily resolve the type. Internally, we're using `typing.get_type_hints`, where by default the "appropriate namespace dictionaries are inferred from obj" when selecting the global and local namespaces used for type lookup. The issue is if for instance you have some inner defined type


```python
def my_func(t):

    @dataclass
    class MyDataclass:
        a : t
```

where `t` itself is not part of the global namespace, then `typing.get_type_hints(MyDataclass)` won't be able to resolve the type `t` because it's not part of its local namespace either.

In short, at least for the purposes of Pytest, I'm not sure we can really modify the tests to accommodate for this use case because the tests are exactly setup like this where the types are passed in arguments.

In practice, I'm not too sure how big a problem this is because as I understand, the main motivation was when using Mitsuba where we can define dataclasses containing fields with alias types (`mi.Float`) before the variant has been set. And if you're adding Mitsuba to the global namespace then it should be fine.

Anyway, I hope that's not too much of an annoyance but please let me know otherwise.

merlinND:
Hi @rtabbara,

Thanks a lot for checking and the careful explanations.

I understand why `: t` wouldn't work with future-style annotations, it makes sense and I don't think it's a problem.
Actually if I try something like the snippet you shared, Pylance warns me that I'm not supposed to use a variable type in annotations :stuck_out_tongue: 

If we really wanted to add a unit test for future-style annotations support, it could be possible with a separate test file I think, something like:

```py
from __future__ import annotations

@pytest.test_arrays('TODO')
def test01_dataclass_with_postponed_types(t):
    if dr.backend_v(t) == dr.JitBackend.CUDA:
        @dataclass
        class MyDataclass:
            a : dr.cuda.Float16
    elif dr.backend_v(t) == dr.JitBackend.LLVM:
        @dataclass
        class MyDataclass:
            a : dr.llvm.Float16

    ... = dr.full(MyDataclass, 10)
    ... = dr.gather(MyDataclass, ...)
    dr.enable_grad(MyDataclass)
    ...
```

njroussel (approved):
Looks good to me!

PR #307, opened 2024-10-25
call_support: allow override of the Domain

merlinND:
This PR lets users of `DRJIT_CALL_*` macros specify a `domain_` static constexpr method to override the domain name, e.g. based on some of the template parameters.

This is used in this Mitsuba PR: https://github.com/mitsuba-renderer/mitsuba3/pull/1362
If we find an easier way to fix the Mitsuba issue, then this change would no longer be needed.

### DIFF
--- a/include/drjit/call.h
+++ b/include/drjit/call.h
@@ -18,6 +18,22 @@
 
 NAMESPACE_BEGIN(drjit)
 
+NAMESPACE_BEGIN(detail)
+
+template <typename T>
+using has_domain_override = decltype(T::domain_());
+
+template <typename CallSupport>
+constexpr const char *get_domain(const char *fallback) {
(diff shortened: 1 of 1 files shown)


njroussel:
This seems fine, an alternative would be to overload the `DRJIT_CALL_BEGIN` macro to take two arguments: the class name and the domain name. 

Let's avoid the double leading underscore in `__domain()` - it's usually reserved for the compiler :)

merlinND:
> overload the DRJIT_CALL_BEGIN macro to take two arguments: the class name and the domain name.

I thought about that, but since the domain name needs to be computed based on template arguments that are only provided once the `call_support` struct is instantiated, I don't think we could give the domain name as a macro argument.

njroussel:
Oh yes, indeed :/

njroussel (approved):
This looks good to me now, thanks :rocket:

PR #308, opened 2024-11-07
`dr.print` : Fix evaluated mode with active mask

rtabbara:
In the case that the threadstate mask stack is non-empty, we need to combine the passed in mask with the top of the stack before the compress call.


njroussel (approved):
Looks good to me, thanks :+1:

PR #312, opened 2024-11-26
Add a missing #include <string>

alexfh:
`#include <string>` is necessary to allow compiling with
LLVM 20 + libc++.


rtabbara:
Hi @alexfh ,

After discussing internally, I think this change is fine. Sorry for the delay and thanks for the fix!

PR #316, opened 2024-12-02
Fix documentation of lerp

dvicini:
Fixes the docstring of the Python lerp function. Currently, the argument names and docstring are inconsistent.


rtabbara:
Thanks @dvicini !

PR #318, opened 2024-12-10
Fix `dr::gather` of non-jit array with scalar indices

YunHsiao:
This doesn't quite work as intended:

```cpp
using FloatX = drjit::DynamicArray<float>;
using FloatP = drjit::Packet<float, 16>;

drjit::Array<FloatX, 3> v;
drjit::gather<drjit::Array<FloatP, 3>(v, 10); // Should gather indices 10-25 as vector of packets
```

### DIFF
--- a/include/drjit/array_router.h
+++ b/include/drjit/array_router.h
@@ -984,7 +984,9 @@ Target gather(Source &&source, const Index &index, const Mask &mask_ = true,
         static_assert(size_v<Source> == size_v<Target>,
                       "When gathering from a nested array source, the source "
                       "and target types must be compatible!");
-        using Index2 = plain_t<replace_scalar_t<Target, scalar_t<Index>>>;
+        using Index2 = std::conditional_t<is_array_v<Index>,
+            plain_t<replace_scalar_t<Target, scalar_t<Index>>>,
+            plain_t<replace_value_t<Target, Index>>>;
         Target result;
         if constexpr (Target::Size == Dynamic)
             result = empty<Target>(source.size());


rtabbara:
Hi @YunHsiao ,

If it's not too much trouble, would it be possible to add some tests to accompany this change? You can take a look at `TEST_BOTH_FLOAT_AGNOSTIC(05_gather_scalar)` in `tests/mem.cpp` as a guide. Otherwise, if it's a bit too cumbersome, I can help with that.

YunHsiao:
Hi @rtabbara,

I don't think the current test suite covers `drjit::Packet` and `drjit::DynamicArray` at all, as this problem arise only for non-jit-typed arrays. After a quick look into `tests/mem.cpp`, I don't think it's the proper place to add the relevant tests.

rtabbara:
Hi @YunHsiao ,

Oh sorry, I was absent-minded and thinking of `drjit-core`. Currently, we only have Python Dr.Jit tests so we'll have to see what the best way is to test out this code-path. I'll get back asap.

rtabbara:
Hi @YunHsiao ,

I just created a small local test and I think your first changes in lines 987-989 are good, but I don't think the second change in line 1009 is correct

```cpp
size_t offset = index * sizeof(value_t<Source>);
```

I have a simple test

```cpp
    using FloatX = dr::DynamicArray<float>;
    using FloatP = dr::Packet<float, 16>;

    dr::Array<FloatX, 3> v;
    v[0] = dr::arange<FloatX>(1000);
    v[1] = dr::arange<FloatX>(1000);
    v[2] = dr::arange<FloatX>(1000);

    // This is the 10th packet not 10th flat index
    auto y = dr::gather<dr::Array<FloatP, 3>>(v, 10);
   
   // Returning a tuple so I can print it in Python
   return { y[0], y[1], y[2] };
```

the expected output should be 

```
([160, 161, 162, 163, 164, 165, 166, 167, 168, 169, 170, 171, 172, 173, 174, 175], [160, 161, 162, 163, 164, 165, 166, 167, 168, 169, 170, 171, 172, 173, 174, 175], [160, 161, 162, 163, 164, 165, 166, 167, 168, 169, 170, 171, 172, 173, 174, 175])
```

but instead I get

```
([10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25], [10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25], [10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25])
```

When performing a gather, the index is with respect to the target i.e. You're asking for the 10th element of type `dr::Array<FloatP, 3>`

YunHsiao:
> When performing a gather, the index is with respect to the target

Sounds not very reasonable for me... If it's an established convention I can follow suit, but I'd argue that passing source indices is much more natural & useful.

Say in the above example we are dealing with the specific range of the array data, and we do want the source range start from index 10. There's no way to specify this behavior since 10 is not a multiple of the packet size. Besides if I'm running the program with diffenret packet sizes on different platforms, I'd have to convert the indices accordingly. A packet-size-agnostic way would be much nicer.

PR #326, opened 2025-01-13
Set new JIT scopes in PyTorch wrapper methods

Speierers:
PyTorch execute its AD graph traversal on different threads, hence we must be careful creating new JIT scopes in the PyTorch wrapper. This PR addresses this issue.


njroussel (approved):
Thanks :100:

PR #329, opened 2025-01-27
Add PacketOps support for textures

rtabbara:
Texture channel data is packed contiguously, so leverage PacketOps for gathers (and scatters) that show substantial speed-ups (particularly on the LLVM backend).

This is especially true for half-precision textures, where the absence of FP16 SIMD gather instructions results in bloated compiled kernels and a degradation in performance relative to the single-precision counterpart. PacketOps circumvents this issue by avoiding the use of gather intrinsics entirely.

### DIFF
--- a/include/drjit/texture.h
+++ b/include/drjit/texture.h
@@ -17,6 +17,7 @@
 #include <drjit/idiv.h>
 #include <drjit/jit.h>
 #include <drjit/tensor.h>
+#include <drjit/util.h>
 
 #pragma once
 
@@ -56,7 +57,20 @@ public:
     using Int32 = int32_array_t<_Storage>;
     using UInt32 = uint32_array_t<_Storage>;
     using Storage = std::conditional_t<IsDynamic, _Storage, DynamicArray<_Storage>>;
+    using Packet = std::conditional_t<is_jit_v<_Storage>,
+        dr::DynamicArray<_Storage>, _Storage*>;
     using TensorXf = Tensor<Storage>;
+
+    #define DR_TEX_ALLOC_PACKET(name, size)                     \
+        Packet _packet;                                         \
+        _Storage* name;                                         \
+                                                                \
+        if constexpr (is_jit_v<Value>) {                        \
+            _packet = empty<Packet>(m_channels_storage);        \
+            name = _packet.data();                              \
+        } else {                                                \
+            name = (_Storage*)alloca(sizeof(_Storage) * size);  \
+        }
 
     /// Default constructor: create an invalid texture object
     Texture() = default;
@@ -125,14 +139,19 @@ public:
         m_handle = other.m_handle;
         other.m_handle = nullptr;
         m_size = other.m_size;
-        m_shape_opaque = std::move(other.m_shape_opaque);
+        m_channels = other.m_channels;
+        m_channels_storage = other.m_channels_storage;
+        for (size_t i = 0; i < Dimension + 1; ++i)
+            m_shape[i] = std::move(other.m_shape[i]);
         m_value = std::move(other.m_value);
+        m_resolution_opaque = std::move(other.m_resolution_opaque);
         for (size_t i = 0; i < Dimension; ++i)
             m_inv_resolution[i] = std::move(other.m_inv_resolution[i]);
         m_filter_mode = other.m_filter_mode;
         m_wrap_mode = other.m_wrap_mode;
         m_use_accel = other.m_use_accel;
         m_migrated = other.m_migrated;
+        m_tensor_dirty = other.m_tensor_dirty;
     }
 
     Texture &operator=(Texture &&other) noexcept {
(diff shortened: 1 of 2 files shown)


include/drjit/util.h
```diff
@@ -318,4 +318,34 @@ template <typename Value> struct range {
...
+                index.index(),
+                mask.index(), 
+                res_indices,
+                mode);
+
+            for (size_t i = 0; i < packet_size; ++i)
+                out[i] = Value::steal(res_indices[i]);
+            return;
+        }
+    }
```
wjakob, line 344: Minor: Put bottom into `else` block to highlight either/or choice.

include/drjit/texture.h
```diff
@@ -183,36 +207,58 @@ template <typename _Storage, size_t Dimension> class Texture {
...
-        if (value.size() != m_size)
-            jit_raise("Texture::set_value(): unexpected array size!");
 
-        drjit::eval(value);
+        if constexpr (!is_jit_v<_Storage>) {
+            if (value.size() != m_size)
+                jit_raise("Texture::set_value(): unexpected array size!");
+            m_value.array() = value;
+        } else /* JIT variant */ {
+
```
wjakob, line 216: very minor: remove newline

include/drjit/texture.h
```diff
@@ -183,36 +207,58 @@ template <typename _Storage, size_t Dimension> class Texture {
...
-        if constexpr (HasCudaTexture) {
-            if (m_use_accel) {
-                value.eval_(); // Sync the value before copying to texture memory
+            if (padded_value.size() != m_size)
+                jit_raise("Texture::set_value(): unexpected array size!");
 
-                size_t tex_shape[Dimension + 1];
-                reverse_tensor_shape(tex_shape, true);
-                jit_cuda_tex_memcpy_d2t(Dimension, tex_shape, value.data(), m_handle);
+            drjit::eval(padded_value);
```
wjakob, line 232: Is that redundant wrt. the `eval_` below?

include/drjit/texture.h
```diff
@@ -282,25 +328,31 @@ template <typename _Storage, size_t Dimension> class Texture {
...
     const TensorXf &tensor() const {
-        if constexpr (HasCudaTexture) {
-            if (m_use_accel && m_migrated) {
-                Storage primal = empty<Storage>(m_size);
 
-                size_t tex_shape[Dimension + 1];
-                reverse_tensor_shape(tex_shape, true);
-                jit_cuda_tex_memcpy_t2d(Dimension, tex_shape, m_handle,
-                                        primal.data());
+        if constexpr (!is_jit_v<_Storage>)
```
wjakob, line 332: minor: put braces since the rest of the `if/else` sequence uses them.

tests/test_texture.py
```diff
@@ -686,7 +686,7 @@ def test22_fetch_grad(t, texture_type):
     assert dr.allclose(4, out[3][0])
 
     for i in range(0, 4):
-        dr.backward(out[i][0])
+        dr.backward(out[i][0], flags=dr.ADFlag.ClearVertices)
```
wjakob, line 689: Is that change intentional?

By the way, does `test_texture.py` test both 3 and 4-channel textures with padding and un-padding? If not, then that would be something to add.

wjakob (commented):
This looks great, only a few quite superficial comments 🚀

include/drjit/texture.h
```diff
@@ -327,28 +375,32 @@ template <typename _Storage, size_t Dimension> class Texture {
...
                                     active.index(), out_idx);
 
-                for (size_t ch = 0; ch < channels; ++ch)
-                    out[ch] = Value(Float32::steal(out_idx[ch]));
+                for (size_t ch = 0; ch < m_channels_storage; ++ch) {
+                    Float32 v = Float32::steal(out_idx[ch]);
+
+                    if (ch < m_channels)
+                        out[ch] = Value(v);
+                }
```
njroussel, line 397:
```suggestion
                for (size_t ch = 0; ch < m_channels; ++ch)
                    out[ch] = Value(Float32::steal(out_idx[ch]));
```

We're just skipping the cases where `ch >= m_channels_storage`, might as well not do them at all, no?

include/drjit/texture.h
```diff
@@ -510,24 +560,31 @@ template <typename _Storage, size_t Dimension> class Texture {
...
+                                v2 = Float32::steal(out_idx[ch*4 + 1]),
+                                v3 = Float32::steal(out_idx[ch*4 + 2]),
+                                v4 = Float32::steal(out_idx[ch*4 + 3]);
+
+                        if (ch < m_channels) {
+                            out[2][ch] = v1;
+                            out[3][ch] = v2;
+                            out[1][ch] = v3;
+                            out[0][ch] = v4;
+                        }
```
njroussel, line 582: Same as above: we could just iterate until `m_channels` to get rid of the `if` condition.

include/drjit/texture.h
```diff
@@ -282,25 +325,30 @@ template <typename _Storage, size_t Dimension> class Texture {
...
+                        Dimension + 1, m_shape);
+                }
+                m_tensor_dirty = false;
             }
-        }
 
-        return m_value;
+            return m_unpadded_value;
+        }
     }
```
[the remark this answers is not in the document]
njroussel: Ignore this, I didn't see the call to `sync_host_data`

include/drjit/texture.h
```diff
@@ -183,36 +207,55 @@ template <typename _Storage, size_t Dimension> class Texture {
...
-                jit_cuda_tex_memcpy_d2t(Dimension, tex_shape, value.data(), m_handle);
+            if constexpr (HasCudaTexture) {
+                if (m_use_accel) {
+                    padded_value.eval_(); // Sync the value before copying to texture memory
 
-                if (migrate) {
-                    Storage dummy = zeros<Storage>(m_size);
+                    size_t tex_shape[Dimension + 1];
+                    reverse_tensor_shape(tex_shape, true);
+                    jit_cuda_tex_memcpy_d2t(Dimension, tex_shape,
```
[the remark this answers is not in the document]
rtabbara: We definitely could, but my understanding was that we wanted to simplify tex creation in `drjit-core` by guaranteeing a channel count that is pow 2 (>= 4)? Maybe that precondition is overkill?

include/drjit/texture.h
rtabbara: I originally had this, but realised we're leaking JIT vars. You still want to steal the variables between `[m_channels, m_channels_storage]` so that we dec ref the underlying JIT variables

include/drjit/texture.h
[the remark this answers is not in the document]
wjakob: That's right. If we do padding and un-padding here, we don't need to repeat that logic in `drjit-core`.

njroussel (approved):
LGTM ! :rocket:

PR #336, opened 2025-02-03
Frozen Functions

DoeringChristian:
This PR implements the frozen function feature in `drjit`.
It adds a new `FrozenFunction` struct that is a wrapper over a Python callable.
When calling the `operator()` of the object, it traverses the input and determines a unique layout/key.
This key is then used to determine if the function has to be recorded or if it's possible to replay it.
The PR relies on functionality implemented in [Frozen ThreadState](https://github.com/mitsuba-renderer/drjit-core/pull/107), and #288.

This replaces #317 and changes the source branch.


include/drjit/array_traverse.h
```diff
@@ -198,14 +218,36 @@ void traverse_1_fn_ro(const Value &value, void *payload, void (*fn)(void *, uint
...
+        for (auto elem : value) {
+            traverse_1_fn_ro(elem, payload, fn);
+        }
+    } else if constexpr (is_detected_v<detail::det_const_get, Value>) {
+        const auto *tmp = value.get();
+        traverse_1_fn_ro(tmp, payload, fn);
+    } else if constexpr (is_detected_v<detail::det_traverse_1_cb_ro, Value *>) {
+        value.traverse_1_cb_ro(payload, fn);
+    } else {
+        // static_assert(false, "Failed to traverse field!");
```
njroussel, line 233: I think it would be too constraining if we raised an error here (same for the `_rw` function).

include/drjit/python.h
```diff
@@ -1057,25 +1060,74 @@ template <typename T> void bind_all(ArrayBinding &b) {
...
+inline void traverse_py_cb_ro(const TraversableBase *base, void *payload,
+                              void (*fn)(void *, uint64_t, const char *variant,
+                                         const char *domain)) {
+    namespace nb    = nanobind;
+    nb::handle self = base->self_py();
+    if (!self)
+        return;
+
+    auto detail = nb::module_::import_("drjit.detail");
+    nb::callable traverse_py_cb_ro =
```
njroussel, line 1103: I'm surprised that the compiler isn't annoyed at the reuse of "traverse_py_cb_ro" here. Maybe for clarity it's worth renaming anyway.

include/drjit/python.h
```diff
@@ -54,6 +54,9 @@
 #include <drjit/math.h>
 #include <drjit-core/python.h>
 #include <nanobind/stl/array.h>
+#include <nanobind/intrusive/counter.h>
+#include <nanobind/nanobind.h>
```
njroussel, line 58: Are these two needed?

njroussel (commented):
This is looking good so far, I still have to go through some of the main bits like `freeze.cpp`/`freeze.h`. I'll come back to this tomorrow.

src/python/common.h
```diff
@@ -89,6 +90,15 @@ inline nb::object get_dataclass_fields(nb::handle tp) {
     return result;
 }
 
+inline drjit::TraversableBase *get_traversable_base(nb::handle h){
+    nb::handle tp = h.type();
+    if (nb::hasattr(tp, DR_STR(_traverse_1_cb_ro)) &&
+        nb::hasattr(tp, DR_STR(_traverse_1_cb_rw))) {
```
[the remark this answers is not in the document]
DoeringChristian: I agree, this is a bit hacky. But I'm not sure if I understand. How would you expose the C++ hierarchy?
Alternatively, is there a way to dynamic cast to a C++ base class from a nanobind object?

tests/test_freeze.py
```diff
@@ -0,0 +1,2636 @@
...
+    assert dr.all(t(2, 2, 2) == o0)
+
+    dr.flush_kernel_cache()
+
+    i0 = t(1, 2, 3)
+    i1 = t(3, 2, 1)
+
+    # Flushing the kernel cache should force a re-trace
+    o0 = func(i0, i1)
+
```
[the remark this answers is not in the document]
DoeringChristian: For context, I added this to test that I'm checking if the kernel cache is checked before replaying. If this would not be the case, this could cause segfaults. Adding a test to the output makes sense.

tests/test_freeze.py
```diff
@@ -0,0 +1,2636 @@
+import drjit as dr
```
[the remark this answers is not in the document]
DoeringChristian: Fair. I also haven't removed a lot of comments yet.
Some of the tests I have ported from Merlin's implementation, especially the more extensive ones.

include/drjit/traversable_base.h
```diff
@@ -0,0 +1,246 @@
...
+                          drjit::detail::traverse_callback_ro fn)              \
+        const override {                                                       \
+        if constexpr (!std ::is_same_v<Base, drjit ::TraversableBase>)         \
+            Base ::traverse_1_cb_ro(payload, fn);                              \
+        drjit::traverse_py_cb_ro(this, payload, fn);                           \
+    }                                                                          \
+    void traverse_1_cb_rw(void *payload,                                       \
+                          drjit::detail::traverse_callback_rw fn) override {   \
+        if constexpr (!std ::is_same_v<Base, drjit ::TraversableBase>)         \
+            Base ::traverse_1_cb_rw(payload, fn);                              \
```
njroussel, line 210:
```suggestion
            Base::traverse_1_cb_rw(payload, fn);                              \
```

include/drjit/traversable_base.h
```diff
@@ -0,0 +1,246 @@
...
+ * class, allows for the automatic traversal of all python members in any
+ * derived python class.
+ */
+#define DR_TRAMPOLINE_TRAVERSE_CB(Base)                                        \
+public:                                                                        \
+    void traverse_1_cb_ro(void *payload,                                       \
+                          drjit::detail::traverse_callback_ro fn)              \
+        const override {                                                       \
+        if constexpr (!std ::is_same_v<Base, drjit ::TraversableBase>)         \
+            Base ::traverse_1_cb_ro(payload, fn);                              \
```
njroussel, line 204:
```suggestion
            Base::traverse_1_cb_ro(payload, fn);                              \
```

njroussel (commented):
I've gone through the last few files I had missed. 

Ignoring the few new comments, I'm overall pretty happy with the current state of the source code. As for the documentation, the only part that truly matters is the `dr.freeze` and I think it's good too - I don't really have the full picture yet, so take that with a pinch of salt. For the C++ documentation, it was plenty enough for me to understand how these mechanisms work and should be used :+1:

src/python/freeze.cpp
```diff
@@ -0,0 +1,1570 @@
...
+                repr << *key;
+
+                std::ostringstream repr_prev;
+                repr_prev << *prev_key;
+
+                jit_log(LogLevel::Debug, "new key: %s", repr.str().c_str());
+                jit_log(LogLevel::Debug, "old key: %s",
+                        repr_prev.str().c_str());
+            }
+#endif
```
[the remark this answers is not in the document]
DoeringChristian: I thinks so, it prints the "layout"/"signature" of the function if it has changed, and might be useful to debug issues where this changed (for example, because the user increments a python integer every time they call the frozen function).

src/python/common.h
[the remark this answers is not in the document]
DoeringChristian: Ah, didn't know it's possible to do that with multiple classes in the inheritance tree.  Currently, I'm ensuring in `bind_traverse`, which is the function that adds the `_traverse_1_cb_ro/rw` methods, that the class inherits from `TraversableBase`.
```cpp
// Expose already existing object tree traversal callbacks (T::traverse_1_..) in Python.
// This functionality is needed to traverse custom/opaque C++ classes and correctly
// update their members when they are used in vectorized loops, function calls, etc.
template <typename T, typename... Args> auto &bind_traverse(nanobind::class_<T, Args...> &cls)
{
    namespace nb = nanobind;
    struct Payload {
        nb::callable c;
    };

    static_assert(std::is_base_of_v<TraversableBase, T>);

    cls.def("_traverse_1_cb_ro", [](T *self, nb::callable c) {
        Payload payload{ std::move(c) };
        self->traverse_1_cb_ro((void *) &payload,
                               [](void *p, uint64_t index, const char *variant,
                                  const char *domain) {
                                   ((Payload *) p)->c(index, variant, domain);
                               });
    });

    cls.def("_traverse_1_cb_rw", [](T *self, nb::callable c) {
        Payload payload{ std::move(c) };
        self->traverse_1_cb_rw((void *) &payload, [](void *p, uint64_t index,
                                                     const char *variant,
                                                     const char *domain) {
            return nb::cast<uint64_t>(
                ((Payload *) p)->c(index, variant, domain));
        });
    });

    return cls;
}
```

drjit/__init__.py
```diff
@@ -1878,6 +1879,142 @@ def binary_search(start, end, pred):
...
+                closure = inspect.getclosurevars(f)
+                self.closure = (closure.nonlocals, closure.globals)
+                self.frozen = detail.FrozenFunction(inner)
+
+            def __call__(self, *args, **kwargs):
+                _state = state(*args, **kwargs) if state is not None else None
+                return self.frozen([self.closure, _state], *args, **kwargs)
+
+            @property
+            def n_recordings(self):
```
merlinND, line 1975: Add docstrings to the methods that are intended to be used by the users.

drjit/__init__.py
```diff
@@ -1878,6 +1879,142 @@ def binary_search(start, end, pred):
...
+                return self.frozen.n_recordings
+
+            @property
+            def n_cached_recordings(self):
```
merlinND, line 1979: What's the difference between `n_recordings` and `n_cached_recordings`? (Should also be explained in the docstrings)

drjit/__init__.py
```diff
@@ -1878,6 +1879,142 @@ def binary_search(start, end, pred):
...
+S = TypeVar("S")
+# Represents the frozen function passed to the decorator without arguments
+F = TypeVar("F")
+# Represents the frozen function passed to the decorator with arguments
+F2 = TypeVar("F2")
+
+@overload
+def freeze(f: None = None, *, state: Callable[..., S]) -> Callable[[F], F]:
+    """
+    Decorator to freeze a function for replaying kernels without compilation.
```
merlinND, line 1892:
```suggestion
    Decorator to freeze a function for replaying kernels without re-tracing.
```

drjit/__init__.py
```diff
@@ -1878,6 +1879,142 @@ def binary_search(start, end, pred):
...
+    kernels. The frozen function keeps a record of all previously recorded calls,
+    and associates them with the input layout. Since the kernels might be compiled
+    differently when changing certain attributes of variables, the frozen function
+    will be re-recorded if the inputs are no longer compatible with previous
+    recordings. We also track the layout of any container and the values of Python
+    types of the inputs to the frozen function. The following are the input
+    attributes which are tracked and might cause the function to be re-recorded if
+    changed.
+
+    - Python type of any variable or container
```
merlinND, line 1909:
```suggestion
    - Type of any variable or container
```

drjit/__init__.py
```diff
@@ -1878,6 +1879,142 @@ def binary_search(start, end, pred):
...
+
+    The width of a variable itself is not tracked, as we want to allow replaying of
+    kernels with different sizes. However, if multiple variables are scheduled and
+    evaluated at the same time, they can get compiled into the same kernel if their
+    size matches. Therefore, we track the size class of the variable, and re-record
+    the kernel if it changed. Because the size is not tracked it is leaked into the
+    frozen function context and we infer the size of kernels using a heuristic based
+    on the sizes of the input variables. This allows gathering from variables with
+    their width as an argument. However, using the size of a variable in more
+    complicated computations might lead to undefined behavior.
```
merlinND, line 1928:
```suggestion
    Generally, the widths of the DrJit variables in the input can be changed in
    each call without triggering a retracing of the function.
    However, variables that had the same width during recording must keep the
    same width in future calls, otherwise the function must be traced again.
    
    Please be particularly wary of using `dr.width()` within a recorded function.
    Usage such as `dr.arange(UInt32, dr.width(a))` is allowed, since `dr.arange`
    uses the width only implicitly, via the launch dimension of the kernel.
    But direct usage, or computations on the width, would no longer be correct
    in subsequent launches of the frozen function if the width of the inputs changes,
    for example:
```

merlinND (commented):
Very partial review, I only looked at the files that had < ~150 lines of diff.

include/drjit/texture.h
```diff
@@ -1494,11 +1495,14 @@ template <typename _Storage, size_t Dimension> class Texture {
...
     bool m_use_accel = false;
     mutable bool m_migrated = false;        /* CUDA backend flag to indicate
-                                               whether texture data is 
+                                               whether texture data is
                                                exclusively on the device */
     mutable bool m_tensor_dirty = false;    /* Flag to indicate whether
                                                public-facing unpadded tensor
                                                needs to be updated */
+
+    DR_TRAVERSE_CB(drjit::TraversableBase, m_value, m_resolution_opaque,
```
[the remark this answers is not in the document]
DoeringChristian: It hasn't been so far, but I'll take a look at how it is used.

src/python/apply.h
```diff
@@ -83,6 +84,9 @@ struct TransformCallback {
     // Type-erased form which is needed in some cases to traverse into opaque
     // C++ code. This one just gets called with Jit/AD variable indices, an
     // associated Python/ instance/type is not available.
+    // This can optionally return a non-owning jit_index, that will be assigned
+    // to the underlying variable if \c traverse is called with the \c rw
+    // argument set to \c true.
```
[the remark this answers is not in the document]
DoeringChristian: Actually, I noticed with the new way of scheduling variables, this change would no longer be required. I'll try to remove it.

src/python/detail.cpp
```diff
@@ -281,6 +283,63 @@ bool leak_warnings() {
...
+        nb::callable m_callback;
+
+        PyTraverseCallback(nb::callable c) : m_callback(c) {}
+    };
+
+    PyTraverseCallback traverse_cb(std::move(c));
+
+    auto dict = nb::borrow<nb::dict>(nb::getattr(self, "__dict__"));
+
+    for (auto value : dict.values()) {
```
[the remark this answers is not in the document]
DoeringChristian: Good point. I haven't looked into that possibility.

drjit/__init__.py
```diff
@@ -1878,6 +1879,142 @@ def binary_search(start, end, pred):
...
+
+    ```python
+    y = dr.gather(type(x), x, dr.width(x)//2)
+    ```
+
+    Similarly, calculating the mean of a variable relies on the number of entries,
+    which will be baked into the frozen function. To avoid this, we suggest
+    supplying the number of entries as a Dr.Jit literal in the arguments to the
+    function.
```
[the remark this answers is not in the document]
DoeringChristian: Unfortunately I have not yet been able to implement the symbolic width yet, so we still have to use this workaround.

include/drjit/texture.h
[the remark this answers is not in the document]
DoeringChristian: I think I fixed this. It required some changes to the cuda texture in `drjit-core` to be able to traverse it's indices. I'm using a similar approach here to what I've implemented for the traversal of the acceleration structure on llvm/embree.

tests/test_custom_type_ext.py
```diff
@@ -69,3 +68,67 @@ def test03_cpp_make_opaque(t):
...
+
+    a = pkg.CustomA(v)
+    assert dr.detail.collect_indices(a) == [v.index]
+
+
+@pytest.test_arrays("float32,-diff,shape=(*),jit")
+def test05_traverse_py(t):
+    """
+    Tests the implementation of ``traverse_py_cb_ro``, used for traversal of
+    PyTrees held by C++ objects.
```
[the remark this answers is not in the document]
DoeringChristian: Ah, the comment was not quite correct. It tests that the function used to traverse classes inheriting from traversable base classes. In this case, I just call it directly, to verify just it's functionality.

src/python/detail.cpp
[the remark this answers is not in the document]
DoeringChristian: Added the same kind of recursion guard that `traverse` uses.

src/python/apply.h
```diff
@@ -92,9 +96,25 @@ struct TransformPairCallback {
     virtual void operator()(nb::handle h1, nb::handle h2, nb::handle h3) = 0;
 };
 
-/// Invoke the given callback on leaf elements of the pytree 'h'
-extern void traverse(const char *op, TraverseCallback &callback,
-                     nb::handle h);
+/**
+ * \brief Invoke the given callback on leaf elements of the pytree 'h'
```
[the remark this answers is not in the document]
DoeringChristian: Added this short comment. Do you want it to be more detailed?
```cpp
/**
 * \brief Invoke the given callback on leaf elements of the pytree 'h',
 *     including JIT indices in c++ objects, inheriting from
 *     \c drjit::TraversableBase.
 *
```

src/python/detail.cpp
```diff
@@ -281,6 +283,85 @@ bool leak_warnings() {
...
+        jit_log(LogLevel::Warn, "%s", nb::str(dict).c_str());
```
merlinND, line 335: This log should be removed

src/python/apply.h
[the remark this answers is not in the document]
DoeringChristian: I'm actually inclined to removing most changes in `apply.h` and `apply.cpp`, as they are no longer required.

src/python/common.h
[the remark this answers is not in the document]
DoeringChristian: Implemented through `try_cast` now.

src/python/freeze.cpp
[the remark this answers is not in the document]
DoeringChristian: This now compiles, but it will be replaced in the `auto-opaque` PR with a better implementation #361 .

drjit/__init__.py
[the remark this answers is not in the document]
DoeringChristian: This will be implemented in #361, and I would remove the corresponding text in there.

drjit/__init__.py
```diff
@@ -2401,6 +2401,211 @@ def binary_search(start, end, pred):
...
+def freeze(
+    f: None = None,
+    *,
+    state: Optional[Callable],
+    max_cache_size: Optional[int] = None,
+    warn_recording_count: int = 10,
+    backend: Optional[JitBackend] = None,
+) -> Callable[[F], F]:
+    """
+    Decorator to freeze a function for replaying kernels without re-tracing.
```
wjakob, line 2419: Hi Christian, I did a writing pass over the docstring of the function, here is my suggestion.  I changed the names of some of the decorator parameters to be shorter that will also require corresponding updates elsewhere. Are you OK with those changes??

```
Decorator to "freeze" functions, which improves efficiency by removing
repeated JIT tracing overheads.

In general, Dr.Jit traces computation and then compiles and launches kernels
containing this trace (see the section on :ref:`evaluation <eval>` for
details). While the compilation step can often be skipped via caching, the
tracing cost can still be significant especially when repeatedly evaluating
complex models, e.g., as part of an optimization loop.

The :py:func:`@dr.freeze <drjit.freeze>` decorator adresses this problem by
altogether removing the need to trace repeatedly. For example, consider the
following decorated function:

.. code-block:: python

   @dr.freeze
   def f(x, y, z):
      return ... # Complicated code involving the arguments

Dr.Jit will trace the first call to the decorated function ``f()``, while
collecting additional information regarding the nature of the function's inputs
and regarding the CPU/GPU kernel launches representing the body of ``f()``.

If the function is subsequently called with *compatible* arguments (more on
this below), it will immediately launch the previously made CPU/GPU kernels
without re-tracing, which can substantially improve performance.

When :py:func:`@dr.freeze <drjit.freeze>` detects *incompatibilities* (e.g., ``x``
having a different type compared to the previous call), it will conservatively
re-trace the body and keep track of another potential input configuration.

Frozen functions support arbitrary :ref:`PyTrees <pytrees>` as function
arguments and return values.

The following may trigger re-tracing:

- Changes in the **type** of an argument or :ref:`PyTree <pytrees>` element.
- Changes in the **length** of a container (``list``, ``tuple``, ``dict``).
- Changes of **dictionary keys**  or **field names** of dataclasses.
- Changes in the AD status (:py:`dr.grad_enabled() <drjit.grad_enabled>`) of a variable.
- Changes of (non-PyTree) **Python objects**, as detected by mismatching ``id()``.

The following more technical conditions also trigger re-tracing:
- A Dr.Jit variable changes from/to a **scalar** configuration (size ``1``).
- The sets of variables of the same size change. In the example above, this
  would be the case if ``len(x) == len(y)`` in one call, and ``len(x) != len(y)``
  subsequently.
- When Dr.Jit variables reference external memory (e.g. mapped NumPy arrays), the
  memory can be aligned or unaligned. A re-tracing step is needed when this
  status changes.

These all correspond to situations where the generated kernel code may need to
change, and the system conservatively re-traces to ensure correctness.

Frozen functions support arguments with a different variable *width* (see
:py:func:`dr.with() <drjit.width>`) without re-tracing, as long as the sets of
variables of the same width stay consistent.

Some constructions are problematic and should be avoided in frozen functions.

- The function :py:func:`dr.width() <drjit.width>` returns an integer literal
  that may be merged into the generated code. If the frozen function is later
  rerun with differently-sized arguments, the executed kernels will still
  reference the old size. One exception to this rule are constructions like
  `dr.arange(UInt32, dr.width(a))`, where the result only implicitly depends on
  the width value.

- @Christian: Should we say something about calling frozen functions from
  other frozen functions?

- @Christian: Any other caveats? Is the part about dr.mean() still relevant? I
  had removed that part since I thought we had a solution for dr.mean().

**Advanced features**. The :py:func:`@dr.freeze <drjit.freeze>` decorator takes
several optional parameters that are helpful in certain situations.

- **Warning when re-tracing happens too often**: Incompatible arguments trigger
  re-tracing, which can mask issues where *accidentally* incompatible arguments
  keep :py:func:`@dr.freeze <drjit.freeze>` from producing the expected
  performance benefits.

  In such situations, it can be helpful to warn and identify changing
  parameters by name. This feature is enabled and set to ``10`` by default.

   .. code-block:: pycon

      >>> @dr.freeze(warn_after=1)
      >>> def f(x):
      ...     return x
      ...
      >>> f(Int(1))
      >>> f(Float(1))
      >>> <<< @Christian: needs an example of what the warning looks like

- **Limiting memory usage**. Storing kernels for many possible input
  configuration requires device memory, which can become problematic. Set the
  ``limit=`` parameter to enable a LRU cache. This is useful when calls to a
  function are mostly compatible but require occasional re-tracing.

Args:
    limit (Optional[int]): An optional integer specifying the maximum number of
      stored configurations. Once this limit is reached, incompatible calls
      requiring re-tracing will cause the last used configuration to be dropped.

    warn_after (int): When the number of re-tracing steps exceeds this value,
      Dr.Jit will generate a warning that explains which variables changed
      between calls to the function.

    state_fn (Optional[Callable]): This optional callable can specify additional
      state to identifies the configuration. ``state_fn`` will be called with
      the same arguments as that of the decorated function. It should return a
      traversable object (e.g., a list or tuple) that is conceptually treated
      as if it was another input of the function.

    backend (Optional[JitBackend]): If no inputs are given when calling the
      frozen function, the backend used has to be specified using this argument.
      It must match the backend used for computation within the function.
```

DoeringChristian: Regarding
> Changes of (non-PyTree) **Python objects**, as detected by mismatching ``id()``.
I'm keeping track of non-PyTree objects based on a hash I calculate of them. Not sure if taking the `id()` might be better?

DoeringChristian: > - @Christian: Should we say something about calling frozen functions from
  other frozen functions?
  
  This should not be a problem, as the inner function is simply recorded by the outer one. I'll check that we actually test for that though. I think it should however not be something the user has to be concerned about.

DoeringChristian: > - @Christian: Any other caveats? Is the part about dr.mean() still relevant? I
  had removed that part since I thought we had a solution for dr.mean().
  
The `mean` case should be covered by #361 . We could the comment here and remove it in the other PR.

include/drjit/python.h
```diff
@@ -1060,25 +1061,73 @@ template <typename T> void bind_all(ArrayBinding &b) {
...
+        nb::borrow<nb::callable>(nb::getattr(detail, "traverse_py_cb_ro"));
+
+    traverse_py_cb_ro_fn(self,
+        nb::cpp_function([&](uint64_t index, const char *variant,
+                           const char *domain) {
+            fn(payload, index, variant, domain);
+        }));
+}
+
+inline void traverse_py_cb_rw(TraversableBase *base, void *payload,
```
wjakob, line 1113: ditto

include/drjit/array_traverse.h
```diff
@@ -180,10 +194,17 @@ template <typename T> auto labels(const T &v) {
 }
 
 template <typename Value>
-void traverse_1_fn_ro(const Value &value, void *payload, void (*fn)(void *, uint64_t)) {
-    (void) payload; (void) fn;
+void traverse_1_fn_ro(const Value &value, void *payload,
```
wjakob, line 197: This function has over time become quite complex. I think it would be good to add a brief comment saying what kinds of data structures this can traverse.

include/drjit/array_traverse.h
```diff
@@ -198,14 +219,34 @@ void traverse_1_fn_ro(const Value &value, void *payload, void (*fn)(void *, uint
...
+        }
+    } else if constexpr (is_detected_v<detail::det_const_get, Value>) {
+        const auto *tmp = value.get();
+        traverse_1_fn_ro(tmp, payload, fn);
+    } else if constexpr (is_detected_v<detail::det_traverse_1_cb_ro, Value *>) {
+        value.traverse_1_cb_ro(payload, fn);
     }
 }
 
 template <typename Value>
```
wjakob, line 235: ditto

include/drjit/texture.h
```diff
@@ -18,6 +18,8 @@
 #include <drjit/jit.h>
 #include <drjit/tensor.h>
 #include <drjit/util.h>
+#include <drjit/traversable_base.h>
+#include <vector>
```
wjakob, line 22: Can you use drjit::vector? I am trying to keep STL includes outside of core header files.

include/drjit/texture.h
```diff
@@ -1591,6 +1593,48 @@ template <typename Storage_, size_t Dimension> class Texture {
...
+        if (!jit_flag(JitFlag::EnableObjectTraversal))
+            return;
+
+        DRJIT_MAP(DR_TRAVERSE_MEMBER_RO, m_value, m_unpadded_value,
+                  m_resolution_opaque, m_inv_resolution);
+        if constexpr (HasCudaTexture) {
+            uint32_t n_textures = 1 + ((m_channels - 1) / 4);
+            std::vector<uint32_t> indices(n_textures);
+            jit_cuda_tex_get_indices(m_handle, indices.data());
+            for (uint32_t i = 0; i < n_textures; i++) {
```
wjakob, line 1612: can omit braces

include/drjit/texture.h
```diff
@@ -1591,6 +1593,48 @@ template <typename Storage_, size_t Dimension> class Texture {
     mutable bool m_tensor_dirty = false;    /* Flag to indicate whether
                                                public-facing unpadded tensor
                                                needs to be updated */
+
+public:
+    void
+    traverse_1_cb_ro(void *payload,
+                     drjit ::detail ::traverse_callback_ro fn) const override {
+        // Only traverse the texture for frozen functions, since accidentally
```
wjakob, line 1601: let's write: "Traverse the function to react to changes when freezing code via @dr.freeze. In all other contexts, the texture is read-only and does not require traversal" (ditto for traverse_1_cb_rw below)

include/drjit/traversable_base.h
```diff
@@ -0,0 +1,254 @@
...
+#include <drjit/array_traverse.h>
+#include <drjit/map.h>
+#include <nanobind/intrusive/counter.h>
+#include <nanobind/intrusive/ref.h>
+
+NAMESPACE_BEGIN(drjit)
+
+NAMESPACE_BEGIN(detail)
+/**
+ * \brief The callback used to traverse all jit arrays of a C++ object such as a
```
wjakob, line 15: Let's not mention Mitsuba here, Dr.Jit is meant to be more general and generally does not reference it parent project.

wjakob: throughout this file: "jit array" -> "JIT array"

include/drjit/traversable_base.h
```diff
@@ -0,0 +1,254 @@
...
+ *
+ * \return
+ *     The new index of the traversed variable. This index is borrowed, and
+ *     should therefore be non-owning.
+ */
+using traverse_callback_rw = uint64_t (*)(void *payload, uint64_t index,
+                                          const char *variant,
+                                          const char *domain);
+
+inline void log_member_open(bool rw, const char *member) {
```
wjakob, line 64: does it make sense to `#ifndef NDEBUG` the body of this function? Just doing the call to `jit_log` has some cost even if the log level then filters the value.

This is similar to trace-level logging in Dr.Jit-Core (e.g. to check reference counts) that is disabled in non-debug builds. This has a noticeable impact on performance.

src/extra/autodiff.cpp
```diff
@@ -1839,6 +1840,18 @@ void ad_scope_leave(bool process_postponed) {
...
 }
 
+void ad_scope_postponed(drjit::vector<uint32_t> *dst) {
+    LocalState &ls             = local_state;
+    std::vector<Scope> &scopes = ls.scopes;
+    if (scopes.empty())
+        ad_raise("ad_scope_leave(): scope underflow!");
+    Scope &scope = scopes.back();
+
+    for (auto &er : scope.postponed) {
```
wjakob, line 1850: Can omit braces

src/python/apply.cpp
```diff
@@ -604,11 +604,12 @@ struct recursion_guard {
     ~recursion_guard() { recursion_level--; }
 };
 
-void TraverseCallback::operator()(uint64_t) { }
+uint64_t TraverseCallback::operator()(uint64_t, const char *, const char *) { return 0; }
 void TraverseCallback::traverse_unknown(nb::handle) { }
 
 /// Invoke the given callback on leaf elements of the pytree 'h'
-void traverse(const char *op, TraverseCallback &tc, nb::handle h) {
```
wjakob, line 611: Add comment about the meaning of the `rw` parameter (here or in the header file)

src/python/detail.cpp
```diff
@@ -281,6 +283,115 @@ bool leak_warnings() {
...
+    ~recursion_guard() { recursion_level--; }
+};
+} // namespace
+
+/**
+ * \brief Traverses all variables of a python object.
+ *
+ * This function is used to traverse variables of python objects, inheriting
+ * from trampoline classes. This allows the user to freeze for example custom
+ * BSDFs, without having to declare its variables.
```
wjakob, line 308: Can this be written without referencing BSDFs? (a highly rendering specific concept, which seems out of place for Dr.Jit)

src/python/detail.cpp
```diff
@@ -281,6 +283,115 @@ bool leak_warnings() {
...
+            const ArraySupplement &s = supp(h.type());
+            auto index_fn = s.index;
+            if (index_fn){
+                if (s.is_class){
+                    auto variant =
+                        nb::borrow<nb::str>(nb::getattr(h, "Variant"));
+                    auto domain =
+                        nb::borrow<nb::str>(nb::getattr(h, "Domain"));
+                    operator()(index_fn(inst_ptr(h)),
+                               variant.is_valid() ? variant.c_str() : "",
```
wjakob, line 324: This `nb::getattr` overload will always return a valid object or raise an exception. So you don't need to call `.is_valid()` below. If you want to support missing attributes, use `nb::getattr(object, "key", nb::handle())`.

src/python/detail.cpp
```diff
@@ -281,6 +283,115 @@ bool leak_warnings() {
...
+            auto index_fn            = s.index;
+            if (index_fn){
+                uint64_t new_index;
+                if (s.is_class) {
+                    auto variant =
+                        nb::borrow<nb::str>(nb::getattr(h, "Variant"));
+                    auto domain = nb::borrow<nb::str>(nb::getattr(h, "Domain"));
+                    new_index   = operator()(
+                        index_fn(inst_ptr(h)),
+                        variant.is_valid() ? variant.c_str() : "",
```
wjakob, line 372: ditto

src/python/detail.cpp
```diff
@@ -281,6 +283,115 @@ bool leak_warnings() {
...
+void traverse_py_cb_ro_impl(nb::handle self, nb::callable c) {
+    recursion_guard guard;
+
+    struct PyTraverseCallback : TraverseCallback {
+        void operator()(nb::handle h) override {
+            const ArraySupplement &s = supp(h.type());
+            auto index_fn = s.index;
+            if (index_fn){
+                if (s.is_class){
+                    auto variant =
```
wjakob, line 319: auto -> nb::str. No harm in being explicit here, the type name is short.

src/python/docstr.rst
```diff
@@ -6056,6 +6056,35 @@
...
+    User code may query this flag to conditionally optimize kernels for frozen
+    function recording, such as re-seeding the sampler, used for rendering.
+
+.. topic:: JitFlag_EnableObjectTraversal
+
+    This flag is set to ``True`` when Dr.Jit is currently traversing
+    inputs and outputs of a frozen function. The flag is automatically managed
+    and should not be updated by application code.
+
+    When enabled, traversal of objects such as the ``Scene`` or ``BSDFs`` in
```
wjakob, line 6085: Mitsuba-specific..

src/python/freeze.h
```diff
@@ -0,0 +1,529 @@
...
+#include <tsl/robin_map.h>
+#include <tsl/robin_set.h>
+
+struct FrozenFunction;
+
+namespace detail {
+
+using index64_vector = drjit::detail::index64_vector;
+
+enum class LayoutFlag : uint32_t {
```
wjakob, line 27: One-line comment saying where this is used and what it describes.

src/python/freeze.h
```diff
@@ -0,0 +1,529 @@
...
+    uint32_t flags = 0;
+
+    /// The literal data
+    uint64_t literal = 0;
+    /// Optional drjit type of the variable
+    VarType vt = VarType::Void;
+
+    /// If a non drjit type is passed as function arguments or result, we simply
+    /// cache it here.
+    /// TODO: possibly do the same for literals?
```
wjakob, line 65: Is TODO still relevant?

src/python/freeze.h
```diff
@@ -0,0 +1,529 @@
...
+ * ``FlatVariables::record_jit_variables``. This struct stores that information
+ * per deduplicated variable.
+ */
+struct VarLayout{
+    /// Optional drjit type of the variable
+    VarType vt = VarType::Void;
+    /// Optional evaluation state of the variable
+    VarState vs = VarState::Invalid;
+    /// Flags, storing information about variables
+    uint32_t flags = 0;
```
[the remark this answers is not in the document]
wjakob: LayoutFlag?

src/python/freeze.h
```diff
@@ -0,0 +1,529 @@
...
+ * This struct stores a flattened representation of a PyTree as well a
+ * representation of it. It can therefore be used to either construct the PyTree
+ * as well as assign the variables to an existing PyTree. Furthermore, this
+ * struct can also be used as a key to the ``RecordingMap``, determining which
+ * recording should be used given an input to a frozen function.
+ * Information about the PyTree is stored in DFS Encoding. Every node of the
+ * tree is represented by a ``Layout`` element in the ``layout`` vector.
+ */
+struct FlatVariables {
+
```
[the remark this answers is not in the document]
wjakob: What kind of flags are stored here?

src/python/freeze.h
```diff
@@ -0,0 +1,529 @@
...
+    void clear() {
+        this->layout_index = 0;
+        this->variables.clear();
+        this->index_to_slot.clear();
+        this->layout.clear();
+        this->backend = JitBackend::None;
+    }
+    /// Borrow all variables held by this struct.
+    void borrow() {
+        for (uint32_t &index : this->variables)
```
wjakob, line 232: ditto

src/python/freeze.h
```diff
@@ -0,0 +1,529 @@
...
+            jit_var_inc_ref(index);
+    }
+    /// Release all variables held by this struct.
+    void release() {
+        for (uint32_t &index : this->variables)
+            jit_var_dec_ref(index);
+    }
+
+    /**
+     * \brief Records information about jit variables, that have been traversed.
```
wjakob, line 242: remove "," in first line

wjakob: jit->JIT

include/drjit/array_traverse.h
```diff
@@ -220,6 +261,16 @@ void traverse_1_fn_rw(Value &value, void *payload, uint64_t (*fn)(void *, uint64
                          is_detected_v<detail::det_traverse_1_cb_rw, Value>) {
         if (value)
             value->traverse_1_cb_rw(payload, fn);
+    } else if constexpr (is_detected_v<detail::det_begin, Value> &&
+                         is_detected_v<detail::det_end, Value>) {
+        for (auto elem : value) {
```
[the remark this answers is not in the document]
DoeringChristian: I'm assuming this refers to the for loop?

src/python/freeze.h
```diff
@@ -0,0 +1,529 @@
...
+     * value of the flat variables will be applied to the ad variable,
+     * preserving the ad_idnex.
+     *
+     * It returns an owning reference.
+     */
+    uint64_t construct_ad_index(uint32_t shrink = 0,
+                                uint64_t prev_index = 0);
+
+    /**
+     * Construct an ad variable given it's layout.
```
wjakob, line 376: ditto: `it's`

wjakob: (I will stop reporting on these, please do a check of all usage in the PR)

src/python/freeze.cpp
```diff
@@ -0,0 +1,1551 @@
...
+    // empty at the end of traversal, we know that no Class variable was
+    // traversed, and registry traversal is not necessary.
+    if (domain && variant && strcmp(domain, "") != 0) {
+        jit_log(LogLevel::Debug, "variant=%s, domain=%s", variant, domain);
+
+        if (domains.empty()){
+            this->variant = variant;
+        }
+        else if (this->variant != variant)
+            jit_raise("traverse(): Variant missmatch! All arguments to a "
```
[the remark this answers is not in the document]
wjakob: typo: previos

src/python/freeze.cpp
```diff
@@ -0,0 +1,1551 @@
...
+    for (uint32_t i = 0; i < var_layout.size(); i++){
+        uint32_t index = variables[i];
+        VarLayout &layout = var_layout[i];
+
+        VarInfo info = jit_set_backend(index);
+
+        if (backend == info.backend || this->backend == JitBackend::None) {
+            backend = info.backend;
+        } else {
+            jit_raise("freeze(): backend missmatch error (backend of this "
```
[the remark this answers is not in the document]
wjakob: will stop reporting, please search and replace throughout the PR

wjakob (commented):
Some comments from me on this PR

tests/test_freeze.py
```diff
@@ -0,0 +1,2865 @@
...
+                reason="The width() of the original input is baked into the kernel to compute the `active` mask during the first launch, so results are incorrect once the width changes."
+            )
+
+        check_results(v, result)
+
+
+@pytest.test_arrays("float32, jit, shape=(*)")
+def test33_multiple_kernels(t):
+    def fn(x: dr.ArrayBase, y: dr.ArrayBase, flag: bool):
+        # TODO: test with gathers and scatters, which is a really important use-case.
```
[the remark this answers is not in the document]
DoeringChristian: I think these are still from Merlin's tests, it looks like they are all covered now (will check them).

src/python/freeze.cpp
```diff
@@ -0,0 +1,1551 @@
+#include "freeze.h"
```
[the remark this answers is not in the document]
DoeringChristian: Will ad that in the auto opaque PR #361 if ok. Since this changes quite a bit.

PR #338, opened 2025-02-06
[AD] Fix missing masks on symbolic calls inside loops

njroussel:
When running AD through evaluated-mode loops, any dynamic dispatch will not have the loop's mask applied.

This PR fixes this issue by saving the loop's mask (top of the mask stack) at the time of tracing the dynamic dispatch into the `CallOp` custom operation. This allows the custom operation to then apply that saved masked whenever it gets traversed.

Without this patch, the issue would typically manifest itself as a CUDA error or segfault during kernel execution. Typically, lanes that should have been masked were accessing out-of-bounds memory.

A regression test was added in `test_switch.py::test18_apply_loop_mask`.

### DIFF
--- a/src/extra/call.cpp
+++ b/src/extra/call.cpp
@@ -581,12 +581,29 @@ public:
         for (size_t i = 0; i < args.size(); ++i)
             m_args.push_back_borrow((uint32_t) args[i]);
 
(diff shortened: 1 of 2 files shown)


rtabbara (approved):
LGTM! :rocket:

PR #339, opened 2025-02-07
Fix #337 Broadcasting of tensors in dr.select

### LINKED ISSUE #337: Broadcasting of tensors in dr.select does not seem to work as before
Ubuntu 24.04, Python 3.12.3, CUDA 12.6, Driver 560.35.03

The following code worked in Dr.Jit 1.0.3 but does not work in 1.0.5 anymore:

```py
import drjit as dr
import numpy as np

# Works
c = dr.cuda.ad.TensorXf(np.zeros([1,2]))
h = dr.cuda.ad.TensorXf(np.ones([1,2]))
dr.select(c==0, 0, h)

# Fails
c = dr.cuda.ad.TensorXf(np.zeros([1,1]))
h = dr.cuda.ad.TensorXf(np.ones([1,2]))
dr.select(c==0, 0, h)
```

As the code snippet shows, it might be related to a broadcasting issue between the input tensors.

rtabbara:
`select` requires promotion of arguments to consistent types including the mask, so we need to handle tensor broadcasting in this case

### DIFF
--- a/src/python/apply.cpp
+++ b/src/python/apply.cpp
@@ -366,14 +366,18 @@ void tensor_broadcast(nb::object &tensor, nb::object &array,
     for (size_t i = 0; i < ndim; ++i)
         dst_size *= shape_dst[i];
 
-    if (src_size == 1) {
-        if (dst_size != 1)
-            array = array + full("zeros", array.type(), nb::int_(0), dst_size);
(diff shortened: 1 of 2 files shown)


wjakob:
Oops, LGTM :)

PR #345, opened 2025-02-18
Optimizer infrastructure

wjakob:
This commit adds standard gradient-based optimizers to the Dr.Jit codebase:

- ``drjit.opt.SGD`` (optionally with momentum)
- ``drjit.opt.RMSProp``
- ``drjit.opt.Adam``

It also adds tests that compare their output to reference data obtained from PyTorch.


drjit/opt.py
```diff
@@ -0,0 +1,1192 @@
...
+        value = dr.detach(value)
+        dr.enable_grad(value)
+
+        if prev is not None and prev[0].shape == value.shape:
+            self.state[key] = value, *prev[1:]
+        else:
+            self._reset(key, value)
+
+    def __len__(self) -> int:
+        """Return the number of number of registered parameters."""
```
Speierers, line 200: typo
```suggestion
        """Return the number of registered parameters."""
```

wjakob: Thanks

wjakob:
I did another pass over the Optimizer API and incorporated the following changes.

- Most of the shared functionality has now been moved to `Optimizer.step()`.
- The optimizer-specific update rules are in a methods of the form `OptimizerSubclass._step`. This is quite a bit less code compared to before.
- The `mask_update` feature is part of `Optimizer`, which means that it is now also available in `SGD`, `RMSProp`, etc.
- We now run `dr.make_opaque()` on anything that is being put into the optimizer.
- Fixed optimization of tensors and added a test.

drjit/opt.py
```diff
@@ -0,0 +1,1199 @@
...
+                    if dr.is_tensor_v(value_tp):
+                        new_value = value_tp(new_value, value.shape)
+                    else:
+                        new_value = value_tp(new_value)
+                dr.enable_grad(new_value)
+
+                # Update the optimizer state and schedule it for evaluation
+                new_state = new_value, lr, new_extra
+
+                dr.schedule(new_state)
```
njroussel, line 451: If `eval == False`, I was almost willing to suggest that we don't even `dr.schedule` the new state, so that users have the full flexibility to decide what gets evaluated.

But it seems like too much of a footgun, users can/should just remove the parameters that they don't want to compute from the optimizer rather than relying on it to not be evaluated.

njroussel (approved):
:rocket:  all good

PR #347, opened 2025-02-20
CustomOp: avoid an unnecessary variable copy

wjakob:
A prior commit (4d71e9c6e2021e994e03203c3646eabd20fd7b8e) changed CustomOp so that it preserves a copy of variables to remember their size. This has been identified as a source of inefficiency in a downstream application. The change in this PR switches the copy to an empty (jit_var_undefined) array of the right size/type, which does not cause an actual allocation/copy to take place.


rtabbara (approved):
Given my initial commit was the culprit I took a look at this. The original motivation for the change was to still be able to fetch the length of the underlying array, so LGTM!

njroussel (approved):
LGTM

PR #348, opened 2025-02-20
Ability to flip the order of axes in Tensor -> nested array conversions

wjakob:
Dr.Jit can convert between nested arrays and tensors (e.g., ``Array3f(TensorXf(...))`` or ``TensorXf(Array3f(...))``), which requires raveling/unraveling the array contents.

This step preserves the shape -- for example
```
x = dr.zeros(Array3f, shape=(3, N))
y = TensorXf(x)
```
returns a tensor of type ``(3, N)``. However, sometimes what is actually needed is shape ``(N, 3)``.

This PR adds a ``flip_axes`` argument to the constructor that accomplishes this.

```
y = TensorXf(x, flip_axes=True)

y = Array3f(y, flip_axes=True)
```

### DIFF
--- a/docs/types.rst
+++ b/docs/types.rst
@@ -517,6 +517,27 @@ each other (e.g. to display a complete 3D vector on each line in the above
 example). In contrast, the string conversion of tensors matches that of other
 array programming libraries and does not transpose the input.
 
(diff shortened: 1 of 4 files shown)


rtabbara (approved):
LGTM! :rocket:

PR #349, opened 2025-02-20
Better handling of unknown objects in ``dr.select()``

wjakob:
The ``dr.select()`` operation is able to traverse PyTrees. Whenever encountering unknown object type, it returns ``None`` in its place.

This commit changes the operation so that it preserves the object if both the true/false branch are consistent. Otherwise, it raises an exception.

### DIFF
--- a/src/python/base.cpp
+++ b/src/python/base.cpp
@@ -901,7 +901,7 @@ nb::object select(nb::handle h0, nb::handle h1, nb::handle h2) {
 
         nb::object transform_unknown(nb::handle h1, nb::handle h2) const override {
             if (!h1.is(h2))
-                nb::raise("encountered incompatible objects with an unknown type (not a Dr.Jit array, not a PyTree).");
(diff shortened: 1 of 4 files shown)


src/python/base.cpp
```diff
@@ -898,8 +898,15 @@ nb::object select(nb::handle h0, nb::handle h1, nb::handle h2) {
     // consistent types so only perform as a last resort
     struct SelectCB : TransformPairCallback {
         SelectCB(nb::handle mask) : mask(mask) {}
+
+        nb::object transform_unknown(nb::handle h1, nb::handle h2) const override {
+            if (!h1.is(h2))
+                nb::raise("encountered incompatible objects with an unknown type (not a Dr.Jit array, not a PyTree).");
```
merlinND, line 904:
```suggestion
                nb::raise("select() encountered incompatible objects with an unknown type (not a Dr.Jit array, not a PyTree).");
```

merlinND (approved):

njroussel (approved):
Looks good to me :+1:

PR #350, opened 2025-02-20
Added ``JitFlag.ForbidSynchronization`` to catch synchronization issues

wjakob:
There can sometimes be lingering issues in a codebase (e.g., forgotten leftovers from a prior debugging session) that trigger costly CPU<->GPU.

The new ``JitFlag.ForbidSynchronization`` flag will raise an exception in ``jitc_sync_thread()`` to immediately identify such problems.


njroussel (approved):
LGTM

PR #353, opened 2025-02-20
Normal random number generation

wjakob:
This commit adds the functions ``PCG32.sample_float32_n`` and ``PCG32.sample_float64_n`` to generate random variates with a standard normal distribution.

On top of this, the commit adds the following bindings

- ``PCG32.sample_float()``
- ``PCG32.sample_float_n()``

They take a type (and optionally, a mask) and then automatically decide between the ``float32`` and ``float64`` variants.


wjakob:
Will do, thank you for the review.

wjakob:
I am merging this despite the CI failure, it's the dreaded texture crash.

PR #354, opened 2025-02-21
Added ``dr.concat()`` operator to concatenate tensors/arrays

wjakob:
This commit adds a new operation to concatenate arrays. It works both for flat 1D arrays and arbitrary tensors.

The name and interface were chosen based on the Array API:

https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.concat.html


njroussel (approved):
LGTM

Just wondering if there was any reason why we wouldn't want to support `axis=None`? It could just be a call to `concat([dr.ravel(i) for i in arr])`, although it wouldn't be the most efficient sequence of scatters...

wjakob:
It's easy to do, I can add this.

PR #358, opened 2025-03-02
``dr.resample()``: new infrastructure to up/downsample arrays/tensors

wjakob:
This commit adds a new function ``dr.resample()`` and underlying infrastructure to resample Dr.Jit arrays/tensors using builtin and custom image reconstruction filters. The operation is differentiable in forward/reverse modes.

For example, the following snippet resizes a single precision tensor using windowed Lanczos filter.

```python
x: TensorXf = ...
y: TensorXf = dr.reshape(x, (10, 20, 30), filter='lanczos')
```
(a faster cubic filter is used when ``filter=..`` is not specified.)

The functionality is equivalent to the
[Image.resize()](<https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.resize>) function of the [Pillow](https://pillow.readthedocs.io/en/stable/) library. The commit adds exhaustive parameterized tests check that the two consistently produce the same result.

The ``drjit.upsample()`` function should no longer be used now that this feature is available, since:

- it does not use a consistent filter when up/downsampling.
- it performs an internal conversion to a texture, which limits the compatible array dimensions.

The function will be removed in a future release and now emits a deprecation warning.


njroussel (approved):
Couple of tiny typos, but the overall logic looks good to me.


I'm not quite sure I understand why we are exporting backed-specific bindings. Couldn't we get away with the generic `drjit::JitArray<JitBackend::None, half/float/double>` here too?

wjakob:
> I'm not quite sure I understand why we are exporting backed-specific bindings. Couldn't we get away with the generic drjit::JitArray<JitBackend::None, half/float/double> here too?

Good point, I switched to generic bindings. Doesn't make the code much prettier, but it is at least consistent with respect to other exports from the extra library.

PR #361, opened 2025-03-07
Frozen Functions improvements

DoeringChristian:
This PR adds a feature to frozen functions, making it possible to not make literal inputs opaque. A mask is computed, that allows us to make literals, changing between function calls to be made opaque. In addition this PR also improves logging and enables freezing of `dr.mean`.


src/python/freeze.cpp
```diff
@@ -615,6 +693,32 @@ void FlatVariables::assign_cb(drjit::TraversableBase *traversable) {
...
+        } else {
+            path += ".";
+            path += suffix;
+        }
+    }
+    scoped_path(std::string &path, uint32_t suffix)
+        : path(path), size(path.size()) {
+        path += "[";
+        path += std::to_string(suffix);
+        path += "]";
```
merlinND, line 717: I think that += on `std::string` is potentially very slow. Replace with an `ostringstream` (?), or simply a large `char *` buffer (in the style drjit-core's buffer used to emit the kernel code).

src/python/freeze.cpp
```diff
@@ -668,50 +780,59 @@ void FlatVariables::traverse(nb::handle h, TraverseContext &ctx) {
                 if (len == DRJIT_DYNAMIC)
                     len = s.len(inst_ptr(h));
 
-                layout.num  = len;
+                layout.num = len;
 
-                for (Py_ssize_t i = 0; i < len; ++i)
+                for (Py_ssize_t i = 0; i < len; ++i){
+                    scoped_path ps(ctx.path, i);
```
merlinND, line 786: Could you please check the performance impact of having string operations for each traverse object?
Having the variable paths is extremely helpful, but we may guard it behind `JitFlag.Debug` or so if it's too expensive.

src/python/freeze.cpp
```diff
@@ -1134,32 +1471,69 @@ size_t
...
+        // struct {
+        //     uint64_t num : 20;
+        //     uint64_t index: 32;
+        //     uint64_t flags: 8;
+        //     uint64_t vt: 4;
+        // };
+        // This makes the assumption that we either don't have more than 1M
+        // elements in one layout.
+        data.push_back(((uint64_t) layout.num << 44) |
+                               ((uint64_t) layout.index << 12) |
```
merlinND, line 1496: Formatting

src/python/freeze.cpp
```diff
@@ -1134,32 +1471,69 @@ size_t
...
+        // therefore we can omit layout.fields.size().
+        // Layout of the hash key data:
+        // struct {
+        //     uint64_t num : 20;
+        //     uint64_t index: 32;
+        //     uint64_t flags: 8;
+        //     uint64_t vt: 4;
+        // };
+        // This makes the assumption that we either don't have more than 1M
+        // elements in one layout.
```
merlinND, line 1494: Could you `assert()` this somewhere?

src/python/freeze.cpp
```diff
@@ -1120,12 +1265,204 @@ void FlatVariables::assign_with_registry(nb::handle dst) {
...
+        return false;
+    }
+
+    uint32_t index = 0;
+    std::string path;
+    log_diff(level, curr, prev, index, path);
+
+    return true;
+}
+inline void hash_combine(uint64_t &seed, uint64_t value) {
```
merlinND, line 1460: Can this be removed if XXH is adopted instead?

src/python/freeze.cpp
```diff
@@ -1216,12 +1594,18 @@ nb::object FunctionRecording::record(nb::callable func,
         // Enter Resume scope, so we can track gradients
         ADScopeContext ad_scope(drjit::ADScope::Resume, 0, nullptr, -1, false);
 
+        jit_log(LogLevel::Info, "traverse output");
```
merlinND, line 1597: Remove leftover log

src/python/freeze.cpp
```diff
@@ -1377,24 +1766,75 @@ nb::object FrozenFunction::operator()(nb::args args, nb::kwargs kwargs) {
...
+            bool new_opaques = false;
+            if (prev_key && auto_opaque)
+                new_opaques =
+                    in_variables->fill_opaque_mask(*prev_key, opaque_mask);
+
+            if (new_opaques) {
+                // If new variables have been discovered that should be made
+                // opaque, we repeat traversal of the input to make them opaque.
+                // This reduces the number of variants that are saved by one.
+                jit_log(LogLevel::Warn,
```
merlinND, line 1822: `Warn` seems a bit strong maybe?

src/python/freeze.cpp
```diff
@@ -1377,24 +1766,75 @@ nb::object FrozenFunction::operator()(nb::args args, nb::kwargs kwargs) {
...
+                        "While traversing the frozen function input, new "
+                        "literal variables have been discovered which changed "
+                        "from one call to another. These will be made opaque, "
+                        "and the input will be traversed again. This will "
+                        "incurr some overhead. To prevent this, make those "
+                        "variables opaque in beforehand. Below, a list of "
```
merlinND, line 1828:
```suggestion
                        "incur some overhead. To prevent this, make those "
                        "variables opaque beforehand. Below, a list of "
```

tests/test_freeze.py
```diff
@@ -70,12 +72,13 @@ def func(x, y):
...
 @pytest.test_arrays("uint32, jit, shape=(*)")
-def test03_output_tuple(t):
+@pytest.mark.parametrize("auto_opaque", [False, True])
+def test03_output_tuple(t, auto_opaque):
     """
     Tests that returning tuples from frozen functions is possible.
     """
 
-    @dr.freeze
+    @dr.freeze(auto_opaque = auto_opaque)
```
merlinND, line 81:
```suggestion
    @dr.freeze(auto_opaque=auto_opaque)
```

and same everywhere below

tests/test_freeze.py
```diff
@@ -2797,8 +2862,74 @@ def func():
...
+def test74_auto_opaque_retraverse(t):
+
+    def func(x: t):
+        return x + 1
+
+    frozen = dr.freeze(func, auto_opaque=True)
+
+    n = 3
+    for i in range(n):
+        x = t(i)
```
merlinND, line 2894:
```suggestion
        x = t(i)  # Create as literal
```

tests/test_freeze.py
```diff
@@ -2797,8 +2862,74 @@ def func():
...
+
+        res = frozen(x)
+        ref = func(x)
```
merlinND, line 2897: Add some asserts about the status of `x` (literal / evaluated) based on `i`. I think it should be evaluated after calling `frozen(x)` when `i >= 1`, and literal otherwise?

tests/test_freeze.py
```diff
@@ -2797,8 +2862,74 @@ def func():
...
+
+    frozen = dr.freeze(func, warn_after=3, auto_opaque=auto_opaque)
+
+    n = 10
+    for i in range(n):
+        lit = dr.zeros(t, (i+1) * 10)
+        x = lit + 0.5
+        dr.make_opaque(x)
+
+        res = frozen(x, lit)
```
merlinND, line 2919: Same as above (assert literal / evaluated)

tests/test_freeze.py
```diff
@@ -2853,13 +2988,383 @@ def init_optimizer():
...
+    Float32 = mod.Float32
+    TensorXf = mod.TensorXf
+
+    def func(x):
+        return dr.mean(x)
+
+    frozen = dr.freeze(func, auto_opaque=auto_opaque)
+
+    for i in range(3):
+        shape = ((i + 3), 10)
```
merlinND, line 3091: Test 3D tensor, and also test that an incompatible shape _does_ trigger a retrace

tests/test_freeze.py
```diff
@@ -2853,13 +2988,383 @@ def init_optimizer():
...
+        func(ref, i, 1)
+
+        assert res[i] == ref[i]
+
+
+@pytest.test_arrays("float32, jit, shape=(*)")
+@pytest.mark.parametrize("auto_opaque", [False, True])
+def test80_tensor_mean(t, auto_opaque):
+    """
+    Tests that the mean of a tensor inside a frozen funciton is computed correctly
```
merlinND, line 3078:
```suggestion
    Tests that the mean of a tensor inside a frozen function is computed correctly
```

tests/test_freeze.py
```diff
@@ -2853,13 +2988,383 @@ def init_optimizer():
...
+@pytest.test_arrays("float32, jit, diff, shape=(*)")
+@pytest.mark.parametrize("auto_opaque", [False, True])
+def test87_tensor_indexing(t, auto_opaque):
+    """
+    Tests that changes in the first dimension of a tensor do not cause re-tracing.
+    """
+    mod = sys.modules[t.__module__]
+
+    def func(x: mod.TensorXf, row: mod.UInt32, col: mod.UInt32):
+        return dr.gather(mod.Float, x.array, row * dr.shape(x)[1] + col)
```
merlinND, line 3293: Use the built-in tensor slicing syntax?  `x[row, col]` or so.

src/python/freeze.cpp
merlinND: Otherwise there's also `std::string::append()`: https://en.cppreference.com/w/cpp/string/basic_string/append.html

src/python/shape.cpp
```diff
@@ -182,6 +184,105 @@ size_t width(nb::handle h) {
...
+            if (value > width)
+                width = value;
+        }
+
+        void traverse_unknown(nb::handle) override {
+            if (width == 0)
+                width = 1;
+            items++;
+        }
+    };
```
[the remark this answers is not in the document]
DoeringChristian: Removed it, as it is not required anymore.

src/python/freeze.cpp
```diff
@@ -639,6 +743,14 @@ void FlatVariables::traverse(nb::handle h, TraverseContext &ctx) {
     try {
         uint32_t layout_index = this->layout.size();
         Layout &layout = this->layout.emplace_back();
+
+        auto it = ctx.visited.find(h.ptr());
+        if (it != ctx.visited.end()) {
+            layout.flags |= (uint32_t) LayoutFlag::RecursiveRef;
+            return;
+        }
+        ctx.visited.insert({ h.ptr(), layout_index });
```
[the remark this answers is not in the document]
DoeringChristian: Thanks for the hint. I changed it to a hashmap and use `try_emplace`, which should only lookup the key once.

PR #363, opened 2025-03-10
Minor: Fix time units in `kernel_history` doc

mworchel:
The docs for `kernel_history` state that the reported times are in microseconds but they seem to be in milliseconds.


njroussel (approved):
Indeed :man_facepalming: 

Thank you!

PR #364, opened 2025-03-12
`dr.replace_grad`: better handling of non-differentiable types and unknown types

njroussel:
This PR changes the semantics of `dr.replace_grad(a, b)`:
* For non-differentiable types, the function call returns `a` rather than the default initialization. For example, if `a = dr.auto.ad.UInt32(0, 1, 2)` then the function now returns `a` instead of `dr.auto.ad.UInt32()` (zero-width value).
* Non unknown types, it will also just return `a` rather than `None`.

This is helpful for Mitsuba records which are `DRJIT_STRUCT`s, where it would result in a ragged value as non-differentiable members of the struct had width 0. The workaround was to exhaustively call `replace_grad` for each differentiable member.

I don't have a strong opinion on how to handle non-Dr.Jit types, it seemed natural to return the unchanged primal value there too.

### DIFF
--- a/src/python/autodiff.cpp
+++ b/src/python/autodiff.cpp
@@ -237,6 +237,11 @@ static void accum_grad(nb::handle target, nb::handle source) {
 
 static nb::object replace_grad(nb::handle h0, nb::handle h1) {
     struct ReplaceGrad : TransformPairCallback {
+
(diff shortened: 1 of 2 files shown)


wjakob:
Looks good to me, thank you!

PR #365, opened 2025-03-14
Slice access/assignment for all Dr.Jit arrays

wjakob:
Slice indexing into (non-tensor) Dr.Jit arrays currently fails with a somewhat infuriating error message:

```text
Complex slicing operations are currently only supported on tensors.
```

This commit adds support for regular arrays to bring them to the same level. In particular, the following are now supported:

- slice access and assignment (potentially nested). For example, the following is now legal for `dr.cuda.Array3f`:

  ```python array[:-1, 1:100:2] = ..`` ```

  These operations are compiled into scatters/gathers.

- Indexing/indexed assignment using integer arrays, e.g.

  ```python index : UInt32 = ... values: Float = ... values[index] = ... ```

- Fully replacing the contents of an array with another array.

  ```python x[:] = y ```


ext/drjit-core
njroussel, line 1: :warning:  This includes all the cooperative vector changes.

ext/nanobind
njroussel, line 1: Are any of these `nanobind` changes required for this PR? I didn't spot anything in the commit history.

I'd prefer to have these merged in a separate commit/PR if it's not the case.

src/python/slice.cpp
```diff
@@ -329,10 +438,15 @@ int mp_ass_subscript(PyObject *self, PyObject *key, PyObject *value) noexcept {
         } else if (key_tp.is(&PyTuple_Type)) {
```
njroussel, line 438: I'm not too sure I understand why the logic in this case was changed. Did the previous assignment logic for tuple indexing set values in a temporary object in some cases?

njroussel (commented):
Core slicing logic looks fine to me!

Just a few notes & questions on some changes.

ext/drjit-core
wjakob: Oops, my bad, will fix.

ext/nanobind
wjakob: My bad, will fix.

src/python/slice.cpp
wjakob: Slice indexing creates new copies of arrays. For example, `x[1:2]` creates a new array. If a slice is used as part of a nested assignment, the previous logic made a change to the temporary instead of the original object. So, at the end, we need to walk backwards and propagate the changes.

wjakob: I will add a comment.

PR #366, opened 2025-03-17
`dr.abs()`: return zero derivative at `x=0`

wjakob:
The derivative of the absolute value function is not defined at x=0. PyTorch returns zero in this case, and Dr.Jit used to return 1. This change makes Dr.Jit consistent with PyTorch.


njroussel (approved):
Looks good to me

PR #367, opened 2025-03-17
Added ``drjit.moveaxis()`` function for rearranging tensor axes

wjakob:
The implementation is consistent with NumPy, and we test against it in a new test. It is differentiable and can move multiple axes at once.


njroussel (approved):
LGTM !

PR #374, opened 2025-03-28
upgrade to nanobind v2.6.0

wjakob:
Nanobind 2.6.0 ships with three improvements that would be nice to have in Dr.Jit / Mitsuba

- Broader ABI compatibility between compilers/compiler versions
- Faster ``nb::getattr()`` function that we extensively use to traverse PyTrees
- Faster function dispatch


njroussel:
(Updating directly to v2.6.1)

PR #377, opened 2025-04-03
DLPack: sync CUDA stream when exporting to TensorFlow

merlinND:
This PR attempts to fix synchronization issues that come up in the unit tests of the new TF interop feature (#301): https://github.com/mitsuba-renderer/drjit/pull/301#issuecomment-2773794399

Since TensorFlow uses non-default CUDA streams for compute and data movement, I believe that we need to synchronize the stream used by DrJit before exporting a tensor to TF.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -621,6 +621,8 @@ Low-level bits
 .. autofunction:: thread_count
 .. autofunction:: set_thread_count
 .. autofunction:: sync_thread
+.. autofunction:: sync_device
+.. autofunction:: sync_all_devices
 .. autofunction:: flush_kernel_cache
 .. autofunction:: flush_malloc_cache
 .. autofunction:: expand_threshold
--- a/src/python/dlpack.cpp
+++ b/src/python/dlpack.cpp
@@ -118,21 +118,24 @@ static nb::ndarray<> dlpack(nb::handle_t<ArrayBase> h, bool force_cpu, nb::handl
                 // https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack__.html
                 /*
                     stream = -1 request producer to perform no synchronization
-                    stream = 0 is ambiguous
+                    stream = 0 is ambiguous (could mean either None, 1, or 2)
                     stream = 1 or None is the legacy default stream
                     stream = 2 is the per-thread default stream
                     stream > 2 is a CUDA handle to the consumer's stream
                 */
                 if (!stream.is_none() && !stream.equal(nb::int_(-1)) && !stream.equal(nb::int_(1))) {
-                    if (stream.equal(nb::int_(0)))
+                    if (stream.equal(nb::int_(0))) {
                         jit_sync_thread();
-                    else {
+                    } else {
+                        // Note: the special value 2 (syncing w.r.t. the per-thread default stream)
+                        // is handled by `jit_cuda_sync_stream()`.
                         uintptr_t stream_handle;
                         if (!nb::try_cast(stream, stream_handle))
                             nb::raise_type_error("__dlpack__(): 'stream' argument must be 'None' or of type 'int'.");
                         jit_cuda_sync_stream(stream_handle);
                     }
                 }
+
             } else {
                 jit_sync_thread();
             }
@@ -265,6 +268,11 @@ void export_dlpack(nb::module_ &) {
       .def("tf",
            [](nb::handle_t<ArrayBase> h) {
                 nb::module_ tf = nb::module_::import_("tensorflow.experimental.dlpack");
-                return tf.attr("from_dlpack")(dlpack(h, false));
+                // TensorFlow uses non-default streams for compute and data transfer, so
+                // we must synchronize on the stream used by DrJit (producer) before
+                // proceeding with TF. Unfortunately, we do not have access to TF's streams,
+                // so we cannot use a lightweight stream-to-stream synchronization.
+                return tf.attr("from_dlpack")(dlpack(h, /* force_cpu */ false,
+                                                     /* stream */ nb::int_(0)));
            }, doc_tf);
 }
--- a/src/python/docstr.rst
+++ b/src/python/docstr.rst
@@ -7144,6 +7144,16 @@
     then you have found a bug. Please report it on the project's
     `GitHub issue tracker <https://github.com/mitsuba-renderer/drjit>`__.
 
+
+.. topic:: sync_device
+
+    Wait for all computation on the current device to finish.
+
+.. topic:: sync_all_devices
+
+    Wait for all computation on *all devices* to finish.
+
+
 .. topic:: flush_malloc_cache
 
     Free the memory allocation cache maintained by Dr.Jit.
--- a/src/python/main.cpp
+++ b/src/python/main.cpp
@@ -171,6 +171,8 @@ NB_MODULE(_drjit_ext, m_) {
     m.def("has_backend", &jit_has_backend, doc_has_backend);
 
     m.def("sync_thread", &jit_sync_thread, doc_sync_thread)
+     .def("sync_device", &jit_sync_device, doc_sync_device)
+     .def("sync_all_devices", &jit_sync_all_devices, doc_sync_all_devices)
      .def("flush_kernel_cache", &jit_flush_kernel_cache, doc_flush_kernel_cache)
      .def("flush_malloc_cache", &jit_flush_malloc_cache, doc_flush_malloc_cache)
      .def("malloc_clear_statistics", &jit_malloc_clear_statistics)


merlinND:
@njroussel After running the relevant unit tests in a loop on my machine for a while, this seems to resolve the synchronization issue. I wouldn't want to introduce unnecessary synchronization points though; do you also think this is the right solution given that TF uses separate non-default streams?

wjakob:
@merlinND, I'm wondering if this change has the potential to affect the conversion of other DLPack tensors negatively? Synchronization is always bad and best avoided. For example, should we perhaps only interpret `2` as a force flush when the thread being active is not the main application thread?

merlinND:
Hi Wenzel,

I agree that we should definitely avoid adding unnecessary synchronizations.

Checking the documentation again, I see that:

>  `stream` is provided by the consumer to the producer to instruct the producer to ensure that operations can safely be performed on the array (e.g., by inserting a dependency between streams via “wait for event”).
> (...)
> - stream = 2 is the per-thread default stream

So essentially, when passing `stream`, the consumer (caller of `__dlpack__()`) indicates which stream they intend to use the value on. So if the `stream` they use is the same stream that we've produced the value on, there's no need for synchronization. Right?

> should we perhaps only interpret 2 as a force flush when the thread being active is not the main application thread?

In DrJit, is the main default stream always used, even when launching kernels from non-main threads?
If so, then `jit_sync_thread()` only syncs the main thread's default stream regardless of which thread it's called from?
Then, indeed we would only want to call `jit_sync_thread()` if `__dlpack__` is being called from a non-main thread.

---

If we make the changes above, then I'm not sure which value to use in order to request a sync when calling `.tf()`, because:

- TF uses non-default streams and doesn't expose any of them, so we can't pass `stream=pointer`
- `.tf()` can be called from the main thread but we always want to sync (because TF never uses the same stream as DrJit AFAICT), so we can't pass `stream=2`
- `stream=0` would currently do the right thing in DrJit, but it's technically disallowed by the spec: "0 is disallowed due to its ambiguity: 0 could mean either None, 1, or 2"

Maybe `stream=0` is still our best bet.

merlinND:
Hi @wjakob,

Could you please check the questions above, I'll update the PR based on your answers and this should fix the flaky TF unit tests.

wjakob:
Hi Merlin,

I just had a moment to take a look at this.

> So essentially, when passing stream, the consumer (caller of __dlpack__()) indicates which stream they intend to use the value on. So if the stream they use is the same stream that we've produced the value on, there's no need for synchronization. Right?

That's right. I looked again what Dr.Jit does, since this changed some time ago. Mitsuba creates a custom stream per device, which is set up with flag `CU_STREAM_DEFAULT`. This means that it synchronizes with respect to the `NULL` stream (global default stream) but not other ones (in particular, it does not implicitly synchronize with respect to per-thread default streams / PTDS). Mitsuba also does _not_ use the PTDS/PTSZ ABI of the cuda driver. We used to, and it caused difficulties with parallel scene loading.
IMO the following behavior is sensible:

- `stream` is -1: user opted out of synchronization, don't do anything.
- `stream` is 0: ambiguous case, synchronize via `jit_sync_thread()`.
- `stream` is 1: user requested to sync with respect to global default thread. Don't need to do anything since we are already synced.
- stream is 2: user requested to sync with respect to the per thread default stream.  We do need to synchronize. But doing a device-wide synchronization that waits on the CPU is a highly pessimistic way of doing that. The better way is to insert an event and wait on that, but it means that we will need to access PTDS/PTSZ-flavored ABI of the cuda driver to get access to the per thread default stream. I've attached an untested patch that tries to do this below, I hope it makes sense.
- `stream` is > 2: we are getting a CUDA stream handle. Pass to `jitc_cuda_sync_stream()`, which will push an event and then wait on that.

```diff
diff --git a/include/drjit-core/jit.h b/include/drjit-core/jit.h
index 720ca16..200d3cc 100644
--- a/include/drjit-core/jit.h
+++ b/include/drjit-core/jit.h
@@ -202,8 +202,11 @@ extern JIT_EXPORT void *jit_cuda_lookup(const char *name);
  * \brief Add CUDA event synchronization between thread state's and external
  * CUDA stream.
  *
- * An event will be recorded into the thread's states stream and the external stream
- * will wait on the event before performing any subsequent work.
+ * An event will be recorded into the thread's states stream and the external
+ * stream will wait on the event before performing any subsequent work. The
+ * special value stream==2 denotes the caller's per-thread default stream.
+ * There is no need to ever synchronize with the global NULL stream, since
+ * Dr.Jit implicitly synchronizes with respect to it.
  *
  * \param stream The CUstream handle of the external stream
  */
diff --git a/src/cuda_api.cpp b/src/cuda_api.cpp
index 293f4de..b3d0f70 100644
--- a/src/cuda_api.cpp
+++ b/src/cuda_api.cpp
@@ -123,6 +123,7 @@ bool jitc_cuda_api_init() {
         LOAD(cuStreamDestroy, "v2");
         LOAD(cuStreamSynchronize);
         LOAD(cuStreamWaitEvent);
+        LOAD(cuStreamWaitEvent_ptsz);
         LOAD(cuPointerGetAttribute);
         LOAD(cuArrayCreate, "v2");
         LOAD(cuArray3DCreate, "v2");
@@ -174,7 +175,7 @@ void jitc_cuda_api_shutdown() {
     Z(cuModuleGetFunction); Z(cuModuleLoadData); Z(cuModuleLoadDataEx); Z(cuModuleUnload);
     Z(cuOccupancyMaxPotentialBlockSize); Z(cuCtxPushCurrent);
     Z(cuCtxPopCurrent); Z(cuStreamCreate); Z(cuStreamDestroy);
-    Z(cuStreamSynchronize); Z(cuStreamWaitEvent); Z(cuPointerGetAttribute);
+    Z(cuStreamSynchronize); Z(cuStreamWaitEvent); Z(cuStreamWaitEvent_ptsz); Z(cuPointerGetAttribute);
     Z(cuArrayCreate); Z(cuArray3DCreate); Z(cuArray3DGetDescriptor);
     Z(cuArrayDestroy); Z(cuTexObjectCreate); Z(cuTexObjectGetResourceDesc);
     Z(cuTexObjectDestroy); Z(cuMemcpy2DAsync); Z(cuMemcpy3DAsync);
diff --git a/src/cuda_api.h b/src/cuda_api.h
index a2a04c0..142d9e5 100644
--- a/src/cuda_api.h
+++ b/src/cuda_api.h
@@ -260,6 +260,7 @@ DR_CUDA_SYM(CUresult (*cuStreamCreate)(CUstream *, unsigned int));
 DR_CUDA_SYM(CUresult (*cuStreamDestroy)(CUstream));
 DR_CUDA_SYM(CUresult (*cuStreamSynchronize)(CUstream));
 DR_CUDA_SYM(CUresult (*cuStreamWaitEvent)(CUstream, CUevent, unsigned int));
+DR_CUDA_SYM(CUresult (*cuStreamWaitEvent_ptsz)(CUstream, CUevent, unsigned int));
 DR_CUDA_SYM(CUresult (*cuMemAllocAsync)(CUdeviceptr *, size_t, CUstream));
 DR_CUDA_SYM(CUresult (*cuMemFreeAsync)(CUdeviceptr, CUstream));

diff --git a/src/cuda_core.cpp b/src/cuda_core.cpp
index 1e54389..cbcbdda 100644
--- a/src/cuda_core.cpp
+++ b/src/cuda_core.cpp
@@ -109,8 +109,12 @@ std::pair<CUmodule, bool> jitc_cuda_compile(const char *buf, bool release_state_
 void jitc_cuda_sync_stream(uintptr_t stream) {
     ThreadState* ts = thread_state(JitBackend::CUDA);
     CUevent sync_event = ts->sync_stream_event;
-    cuda_check(cuEventRecord(sync_event, (CUstream)ts->stream));
-    cuda_check(cuStreamWaitEvent((CUstream)stream, sync_event, CU_EVENT_DEFAULT));
+    scoped_set_context guard(ts->context);
+    cuda_check(cuEventRecord(sync_event, (CUstream) ts->stream));
+    if (stream != 2)
+        cuda_check(cuStreamWaitEvent((CUstream)stream, sync_event, CU_EVENT_DEFAULT));
+    else
+        cuda_check(cuStreamWaitEvent_ptsz(nullptr, sync_event, CU_EVENT_DEFAULT));
 }
 ```

merlinND:
Thank you @wjakob for looking into it and providing the patch! I've opened https://github.com/mitsuba-renderer/drjit-core/pull/139 with your patch.
The solution makes sense to me.

@dvicini, do the changes in this PR make sense to you? In particular, calling `array.tf()` will now trigger a `jit_sync_thread()` call.
The reasoning is that otherwise, the streams used by TF and the stream used by DrJit are not synchronized at all. Did you ever run into sync issues?

wjakob:
@merlinND Is that so? I thought we now insert an event and wait for it asynchronously. That is assuming that TF uses the special `2` argument.

merlinND:
Sorry, it was not very clear because two things were included in this PR:

1. TF uses 3 non-default streams (compute, device-to-host memcpy, host-to-device memcpy) that it creates and manages itself. I didn't find a way to get pointers to those streams, so I don't know how to ensure synchronization w.r.t. TF's streams other than `jit_sync_thread()`.
2. `stream=2` is part of the DLPack spec but was not correctly implemented in DrJit. Your patch added support, which is nice. But I don't think we can use `stream=2` in `array.tf()` because TF doesn't use the per-thread stream.

Please let me know if I missed something. As you said, `jit_sync_thread()` is heavy and we want to avoid it as much as possible.

wjakob:
Ok, that makes sense. Let's wait for @dvicini's on the Google™ viewpoint before merging :-)

merlinND:
Ping @dvicini, do you have an opinion above the above regarding TF interop?
https://github.com/mitsuba-renderer/drjit/pull/377#issuecomment-2809004610

PR #378, opened 2025-04-04
API to convolve arrays/tensors with continuous convolution kernels

wjakob:
This PR adds new function ``drjit.convolve()`` that repurposes the ``drjit.resample()`` infrastructure to convolve one or more axes of a Dr.Jit array or tensor with a 1D filter. The user can choose one of multiple presets or specify a custom function.


njroussel (approved):
LGTM

PR #372, opened 2025-03-24
[doc] Various typo fixes.

SepandKashani:
I went through the v1.0.x docs linearly to update some of my code using drjit v0.4.x. In the process I found a few typos.

This PR only touches documentation.

PR #380, opened 2025-04-08
[dda] Doc fixes

SepandKashani:
The `dda()` example in the [API reference](https://drjit.readthedocs.io/en/stable/reference.html#drjit.dda.dda) does not run due to:
* failed type assertions;
* missing parameter in loop function.

This PR updates the example to use the required integer types and function signature.

### DIFF
--- a/drjit/dda.py
+++ b/drjit/dda.py
@@ -35,7 +35,8 @@ def dda(
        from drjit.scalar import Array3f, Array3u, Float, Bool
 
        def dda_fun(state: list, index: Array3u,
-                   pt_in: Array3f, pt_out: Array3f) -> tuple[list, bool]:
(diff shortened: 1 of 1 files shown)


wjakob:
Thanks, Sepand. Does this need a rebase onto the current version? The PR lists various commits, but only one file is touched at the end.

SepandKashani:
I accidentally made this PR on top of #372, reason why some past (merged) commits are shown again. I can redo the PR fresh to have a single commit if preferable.

merlinND:
Hi @SepandKashani,

If you rebase & squash your branch onto the latest `master`, and then force-push to your branch, this PR will be automatically updated. So there's no need to create another one :)

wjakob:
Thanks!

PR #383, opened 2025-04-13
Make PCG32 sample open interval (0,1)

tomix1024:
Hi,

I've been having problem sampling random numbers in the unit interval multiple times now, so I though to share the solution...

**Problem:**
The integral over a function on a domain (usually the unit hypercube) is only affected by the behaviour of the function in the interior of the domain, **not** the boundary of the domain.
Therefore, when we approximate integrals using Monte-Carlo integration, we want to create random numbers inside of the domain, and not the boundary.
However, the standard `PCG32` implementation samples random numbers in the half-open interval `[0, 1)`, which has caused problems for me multiple times, when the integrand is not well behaved on the boundary.

**Solution:**
Just make sure that the random numbers returned from `PCG32::next_float32()` or `PCG32::next_float64()` are never exactly 0 or 1!
- For 64bit floats, the fix is straight forward. We simply append a 1 to the mantissa after the random bits.
- For 32bit floats, we could simply set the last bit to 1, which would reduce the count of random numbers possible to be generated. *Instead*, we check if the random bits that would usually be used for the mantissa evaluate to 0, and if so, we use the remaining bits of the random integer which would usually be discarded and append 1 to the mantissa like for 64bit integers.

Thanks!

### DIFF
--- a/include/drjit/random.h
+++ b/include/drjit/random.h
@@ -139,18 +139,24 @@ template <typename T> struct PCG32 {
         return Float16(next_float32(mask));
     }
 
-    /// Generate a single precision floating point value on the interval [0, 1)
+    /// Generate a single precision floating point value on the interval (0, 1)
     DRJIT_INLINE Float32 next_float32() {
-        return reinterpret_array<Float32>(sr<9>(next_uint32()) | 0x3f800000u) - 1.f;
+        UInt32 rand_bits = next_uint32();
+        UInt32 mantissa_bits = select((rand_bits & sl<9>(~0u)) != 0u, sr<9>(rand_bits), sl<14>(rand_bits) | sl<13>(1u));
+        UInt32 float_bits = (mantissa_bits) | 0x3f800000u;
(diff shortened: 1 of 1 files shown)


wjakob:
Hi @tomix1024 , the `[0, 1)` half interval is the standard behavior for RNGs and the expected behavior here. If you need different conventions, I suggest that you adapt PCG32 in your project. However, there are no plans to incorporate such behavior here.

tomix1024:
Thanks for the quick response.
Do you happen to know if the standard behaviour is just a convention, or if there are (good) reasons for why `[0, 1)` is used?

wjakob:
Various distributions have some kind of corner case at 1 (which e.g., would result in a point at infinity). In cases that don't fit this pattern, it is easy to adapt the RNG as you have done here.

In any case, this is a standard convention, see e.g., https://docs.python.org/3/library/random.html. Many other RNGs also generate values in this form.

PR #385, opened 2025-04-24
CMake: add option to disable leak warnings

merlinND:
For convenience, it's sometimes helpful to disable leak warnings even in local builds.
The default behavior is preserved: leak warnings are enabled by default, except in PyPi wheel builds.


merlinND:
Hi @wjakob and @njroussel,

Would there be interest in this CMake option to conveniently disable leak warnings in local builds?

njroussel (approved):
LGTM. 

I think this is a decent QOL feature.

PR #390, opened 2025-05-04
Coopvec hashgrid

DoeringChristian:
This PR implements a HashGrid encoding layer for the upcoming neural network interface. It works with the `nn.Sequential` module and can take `CoopVec`, `ArrayXf` or `list`s of `Float` values in both 16 and 32 bit. The output will be either a `CoopVec` of 16 or 32 bit floats depending on the storage type.

The tests still have to be refined, and some of the fields will probably have to be renamed.

The implementation has originally been developed by @merlinND. I have ported it to the upstream branch and extended it to `n` dimensions.

### DIFF
--- a/drjit/nn.py
+++ b/drjit/nn.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 import drjit
 import sys
+from dataclasses import dataclass, field
 
 if sys.version_info < (3, 11):
     from typing_extensions import Tuple, Sequence, Union, Type, TypeAlias, Optional, Any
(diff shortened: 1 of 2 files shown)


drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+    return multiple * div_round_up(num, multiple)
+
+
+@dataclass
+class HashEncodingConfig:
+    """
+    This class holds information about a hash based encoding, such as hash grids
+    or permutohedral encodings.
+
+    Args:
```
merlinND, line 34: For the params below that support it, indicate what special values like `-1` mean.

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+        dimension: The dimensionality of the hash encoding. This corresponds to
+            the number of input features the encoding can take.
+        n_levels: Hash encodings generally make use of multiple levels of the same
+            encoding with different scales. This parameter specifies the number of
+            levels used by this encoding.
+        n_features_per_level: More than one feature can be stored in a vertex per
+            level. This value specifies how many, and the number of output features
+            of the hash encoding layer is given by ``n_levels * n_features_per_level``.
+            This value should always be a multiple of two, in order to ensure efficient
+            gradient backpropagation.
```
merlinND, line 44:
```suggestion
        n_features_per_level: Specifies how many features are stored at each vertex
            and at each level. The number of output features of the hash encoding
            is given by ``n_levels * n_features_per_level``.
            In order to ensure efficient gradient backpropagation, this value should 
            be a multiple of two.
```

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+        self.n_levels = n_levels
+        self.n_features_per_level = n_features_per_level
+        self.hashmap_size = hashmap_size
+        self.base_resolution = base_resolution
+        self.per_level_scale = per_level_scale
+        self.log2_per_level_scale = drjit.log2(per_level_scale)
+        self.align_corners = align_corners
+        self.torchngp_compat = torchngp_compat
+        self.smooth_weight_gradients = smooth_weight_gradients
+        self.smooth_weight_lambda = smooth_weight_lambda
```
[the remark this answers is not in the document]
merlinND: In `__post_init__()`, you can also check that the given parameters are correct, e.g. `per_level_scale >= 1`, etc.
Also you could enforce or warn about non-multiple-of-two `n_features_per_level`.

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+        """
+        Allocates an Encoding with the specified type.
+        Returns the new recording as a separate object instance.
+        """
+
+        result = type(self)()
+        result._config = dataclasses.replace(self._config)
+        result._alloc(dtype)
+
+        return result
```
merlinND, line 118: Is this mechanism needed now that the hashgrid is no longer a NN Layer?
Although it could be convenient to keep the same syntax for all encodings.

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+
+    def n_params(self) -> int:
+        """
+        The number of parameters, held by this encoding.
+        """
+        return drjit.width(self.data)
+
+    def set_params(self, values: drjit.ArrayBase, copy=False) -> None:
+        """
+        Should pass `values` as a DrJit array (float32 only for now).
```
merlinND, line 128: Is this docstring up to date?

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+        """
+        assert drjit.width(values) == drjit.width(self.data)
+        assert type(values) is type(self.data)
+        if copy:
+            self.data = type(self.data)(values)
+        else:
+            assert isinstance(values, type(self.data))
```
merlinND, line 135: Looks redundant with the assert above

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+
+    @property
+    def smooth_weight_lambda(self) -> float:
+        """Blending factor for gradient smoothing using straight-through estimator.
+
+        Controls the strength of the gradient smoothing applied to interpolation weights.
+        A value of 1.0 fully replaces the original gradients with smoothed ones,
+        while 0.0 disables smoothing entirely.
+        """
+        return self._config.smooth_weight_lambda
```
merlinND, line 214: All of these getters could go away if the config object's properties were moved into the class itself.

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+        base_resolution: int = 16,
+        per_level_scale: float = 2,
+        align_corners: bool = False,
+        torchngp_compat: bool = False,
+        smooth_weight_gradients: bool = False,
+        smooth_weight_lambda: float = 1.0,
+    ) -> None: ...
+
+    @overload
+    def __init__(self) -> None: ...
```
merlinND, line 382: Why do we need this if all of the arguments have default values?

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+            torchngp_compat: If this value is ``True``, the encoding will use the
+                same indexing and offset functions compared to tiny-cuda-nn.
+            smooth_weight_gradients: whether to smooth the gradients of the weights
+                by using a straight-through estimator.
+            smooth_weight_lambda: the value of lambda used for the straight-through estimator.
+        """
+        if len(args) == 0 and len(kwargs) == 0:
+            self._config = HashEncodingConfig()
+        else:
+            self._config = HashEncodingConfig(*args, **kwargs)
```
merlinND, line 418: Once that config is moved to the parent class, you should be able to just do:

```py
        super().__init__(*args, **kwargs)
```

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+
+    def __call__(
+        self, p: Iterable[drjit.ArrayBase], active: bool | drjit.ArrayBase = True
+    ) -> Iterable[drjit.ArrayBase]:
+        self._init_types(p)
+
+        p = self.PositionFloatXf(p)
+
+        assert drjit.shape(p)[0] == self.dimension, (
```
merlinND, line 427: Wouldn't `len(p)` give you the outer dimension?

```py
>>> a = [mi.Float([1, 2]), mi.Float([3, 4]), mi.Float([5, 6])]
>>> len(a)
3
>>> b = mi.Point3f(*a)
>>> len(b)
3
```

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+        p = list(p)
+
+        if isinstance(p, list):
+            self.PositionFloat = drjit.leaf_t(p[0])
+        else:
+            self.PositionFloat = drjit.leaf_t(p)
+
+        self.PositionFloatXf = (
+            mod.ArrayXf16 if drjit.is_half_v(self.PositionFloat) else mod.ArrayXf
+        )
```
merlinND, line 253: I don't love that this is done at each call, intuitively evaluating the hashgrid should be `const` (not modify the hashgrid object).
Couldn't these aliases be setup once when initializing / setting the hashgrid parameters? At that point we already know the dimensionality, data precision and backend. It shouldn't change at each call.

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+
+    def _acc_features(
+        self,
+        level_i: int,
+        weight: drjit.ArrayBase,
+        index: drjit.ArrayBase,
+        values: list[drjit.ArrayBase],
```
merlinND, line 260:
```suggestion
        out_values: list[drjit.ArrayBase],
```

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+        active: bool | drjit.ArrayBase,
+    ):
+        """
+        Accumulates the ``self.num_features`` features at the given index.
```
merlinND, line 264:
```suggestion
        Accumulates the ``self.num_features`` features into `values` at the given index.
```

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+        """
+        Given a key i.e. a D-dimensional integer vector, identifying a vertex
+        of a simplex, this function calculates the index of the feature tuple,
+        which is associated with that vertex.
+        """
+        raise NotImplementedError()
+
+    def _level_scale(self, level: int) -> float:
+        """
+        The -1 means that `base_resolution` refers to the number of grid _vertices_ rather
```
merlinND, line 301: This is nice to explain, but an explanation of the function itself should come first in the docstring.

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+            a low enough scale, and use less than ``hashmap_size`` number of parameters
+            per level.
+        base_resolution: The scale factor of the 0th layer in the hash encoding.
+        per_level_scale: To calculate the scale of a layer, the scale of the previous
+            layer is multiplied by this value.
+        align_corners: If this value is ``True``, the simplex vertices are aligned
+            with the domain of the encoding [0, 1].
+        smooth_weight_gradients: whether to smooth the gradients of the weights
+            by using a straight-through estimator.
+        smooth_weight_lambda: the value of lambda used for the straight-through estimator.
```
merlinND, line 351: I guess these would move to the base class' constructor docstring.
(In any case they shouldn't be here since this is the class-level docstring, not the constructor).

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+    """
+
+    DRJIT_STRUCT = {
+        "data": drjit.ArrayBase,
+        "dtype": type,
+        "_config": HashEncodingConfig,
+        "_level_offsets": list,
```
merlinND, line 358: `list[precise type]` if possible, and same below

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+        self._init_types(p)
+
+        p = self.PositionFloatXf(p)
+
+        assert drjit.shape(p)[0] == self.dimension, (
+            f"This hash grid expected an input of feature dimension {self.dimension}"
+            f" but got {drjit.shape(p)[0]}."
+        )
+
+        values = [self.StorageFloat(0.0)] * self.n_features_per_level * self.n_levels
```
merlinND, line 432:
```suggestion
        values = [self.StorageFloat(0.0)] * (self.n_features_per_level * self.n_levels)
```

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+
+        for level_i in range(self.n_levels):
+            scale = self._level_scale(level_i)
+
+            p_offset: float = (
+                0.5 if not self.align_corners or self.torchngp_compat else 0.0
+            )
+            pos = drjit.fma(p, scale, p_offset)
+
```
merlinND, line 441:
```suggestion
```

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+            pos0 = self.ArrayXu(drjit.floor(pos))
+
+            w1 = pos - pos0
+            w0 = 1.0 - w1
+
+            for offset in self._grid_offsets:
+                pos_grid = pos0 + self.ArrayXu(offset)
```
merlinND, line 448: Could the `_grid_offsets` be stored directly as `ArrayXu` to avoid many conversions?
I believe they're initialized at a point where we can already determine what `ArrayXu` is.

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+            self._config = HashEncodingConfig()
+        else:
+            self._config = HashEncodingConfig(*args, **kwargs)
+
+    def __call__(
+        self, p: Iterable[drjit.ArrayBase], active: bool | drjit.ArrayBase = True
+    ) -> Iterable[drjit.ArrayBase]:
+        self._init_types(p)
+
+        p = self.PositionFloatXf(p)
```
merlinND, line 425: Explain that we are keeping the positions and interpolation weights in higher precision

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+                pos_grid = pos0 + self.ArrayXu(offset)
+                weight = drjit.select(self.ArrayXu(offset) == 0, w0, w1)
+                weight = drjit.prod(weight, axis=0)
+
+                index = self.indexing_function(pos_grid, level_i)
+                self._acc_features(level_i, weight, index, values, active)
+
+        values = [v & active for v in values]
+
+        return self.StorageFloatXf(*values)
```
merlinND, line 457:
```suggestion
        return self.StorageFloatXf(*values) & active
```

Should work I think

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+            self.UInt32(2165219737),
+        ]
+
+        # tiny-cuda-nn uses a uint32 here, which can lead to an integer overflow,
+        # and the hash condition not being detected. For compatibility, we use
+        # Dr.Jit's UInt32.
+        stride = self.UInt32(1) if self.torchngp_compat else 1
+        index = self.UInt32(0)
+        for d in range(self.dimension):
+            index += key[d] * self.UInt32(stride)
```
merlinND, line 488: Doesn't this overflow / do something incorrect anyway once you convert to `self.UInt32`, even when `self.torchngp_compat == False`?
If not (maybe because we break out of the loop before), then explain it in the comments because it's not super obvious.

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+        5. Precomputes grid offset patterns for voxel corner indexing
+        """
+        self.dtype = drjit.leaf_t(dtype)
+        self._init_types()
+
+        # TODO: warning for n_features == 3
+
+        assert (
+            self.hashmap_size % 8
+        ) == 0, f"Invalid hashmap size {self.hashmap_size}, must be a multiple of 8."
```
merlinND, line 520:
```suggestion
        assert (self.hashmap_size % 8) == 0, \
               f"Invalid hashmap size {self.hashmap_size}, must be a multiple of 8."
```

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+
+            self._level_offsets[level_i] = offset
+            offset += params_in_level
+
+        self._level_offsets[-1] = offset
+
+        params_size = self._level_offsets[-1] * self.n_features_per_level
+        self.data = drjit.zeros(dtype, params_size)
+
+        # Stores the pattern of offsets used to index the 2** n corners of a voxel
```
merlinND, line 544:
```suggestion
        # Stores the pattern of offsets used to index the 2**n corners of a voxel
```

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+            f"    smooth_weight_lambda={self.smooth_weight_lambda}\n"
+            ")"
+        )
+
+    def level_offset(self, level_i: int) -> int:
+        """
+        Helpful to build e.g. debug visualizations by splitting params per level.
+        Must be called with an index up to and including `n_levels`.
+
+        Warning: the level offset returned does *not* account for the feature count.
```
merlinND, line 572:
```suggestion
        Warning: the level offset is expressed in number of vertices, i.e. it does *not* account for the feature count in each vertex.
```

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+    """
+    Permutohedral lattice-based encoding inspired by :cite:`rosu2023permutosdf`.
+
+    Unlike hash grid encodings that use regular grid lattices, this encoding employs
+    a permutohedral lattice structure where simplices consist of triangles, tetrahedra,
+    and higher-dimensional analogs. The key advantage is linear scaling: the number of
+    vertices per simplex (and thus memory lookups per sample per level) grows linearly
+    with dimensionality, compared to exponential growth in grid-based approaches.
+
+    This implementation simplifies the original method by performing sorting and
```
merlinND, line 589: Here would be a good spot to credit @tszirr

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+    ) -> None: ...
+
+    @overload
+    def __init__(self) -> None: ...
+
+    def __init__(self, *args, **kwargs) -> None:
+        if len(args) == 0 and len(kwargs) == 0:
+            self._config = HashEncodingConfig()
+        else:
+            self._config = HashEncodingConfig(*args, **kwargs)
```
merlinND, line 649: Same comments as bove

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+
+    def __call__(
+        self, p: Iterable[drjit.ArrayBase], active: bool | drjit.ArrayBase = True
+    ) -> Iterable[drjit.ArrayBase]:
+        self._init_types(p)
+
+        p = self.PositionFloatXf(p)
+
+        assert drjit.shape(p)[0] == self.dimension, (
```
merlinND, line 658: Same as above

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+                    xk_i = self.PositionFloat(xk[i])
+                    xk_j = self.PositionFloat(xk[j])
+                    swap = xk_i > xk_j
+                    xk[i] = drjit.select(swap, xk_j, xk_i)
+                    xk[j] = drjit.select(swap, xk_i, xk_j)
+
+                    xki_i = self.Int32(xki[i])
+                    xki_j = self.Int32(xki[j])
+                    xki[i] = drjit.select(swap, xki_j, xki_i)
+                    xki[j] = drjit.select(swap, xki_i, xki_j)
```
merlinND, line 691: @wjakob Would you know of a better way to implement this?
This variant of the hashgrid is specifically designed for `self.dimensions > 3`.

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+
+                pos_grid = base + self.ArrayXu(offset)
+                weight = weights[rank]
+
+                index = self.indexing_function(pos_grid, level_i)
+                self._acc_features(level_i, weight, index, values, active)
+
+        values = [v & active for v in values]
+
+        return self.StorageFloatXf(*values)
```
merlinND, line 727: Same as above

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+            weights = drjit.zeros(
+                self.PositionFloatXf, shape=(self.dimension + 1, drjit.width(p))
+            )
+
+            for rank in range(self.dimension):
+                weights[rank] = xk[rank] - (xk[rank - 1] if rank > 0 else 0)
+                grid_offset = drjit.select(
+                    (fract > xk[rank])
+                    | (fract == xk[rank])
+                    & (self.ArrayXi([i for i in range(self.dimension)]) >= xki[rank]),
```
merlinND, line 704: It would be nice to explain this logic a bit

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+        2. Computes parameter counts per level, using dense grids for small scales
+           and hash tables (limited by hashmap_size) for larger scales
+        3. Calculates memory offsets for each level in the flattened data array
+        4. Allocates the main data storage array
+        5. Precomputes grid offset patterns for voxel corner indexing
+        """
+        self.dtype = drjit.leaf_t(dtype)
+        self._init_types()
+
+        # TODO: warning for n_features == 3
```
merlinND, line 744: I guess this applies to both types of hashgrids, so it should be in the constructor of the base class (assuming you're moving the config parameters there).

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+            params_in_level = next_multiple(params_in_level, 8)
+            params_in_level = min(params_in_level, self.hashmap_size)
+
+            self._level_offsets[level_i] = offset
+            offset += params_in_level
+
+        self._level_offsets[-1] = offset
+
+        params_size = self._level_offsets[-1] * self.n_features_per_level
+        self.data = drjit.zeros(dtype, params_size)
```
merlinND, line 770: Add non-zero init as discussed

drjit/hgrid.py
```diff
@@ -0,0 +1,827 @@
...
+    return (num + divisor - 1) // divisor
+
+
+def next_multiple(num: int, multiple: int) -> int:
+    """Round `num` to the next multiple of `multiple`"""
+    return multiple * div_round_up(num, multiple)
+
+
+@dataclass
+class HashEncodingConfig:
```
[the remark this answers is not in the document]
DoeringChristian: The idea is similar to the "config" field in the `Linear` layer. The hashgrid however has so many parameters, that I thought it would be better to extract it into a dataclass.
From `nn.py`:
```python
    DRJIT_STRUCT = {
        'config': Tuple[int, int, bool],
        'weights': TensorOrViewOrNone,
        'bias': TensorOrViewOrNone
    }

    def __init__(self, in_features: int = -1, out_features: int = -1, bias = True) -> None:
        self.config = (in_features, out_features, bias)
        self.weights = self.bias = None
```

drjit/hgrid.py
DoeringChristian: Most of them could probably. But the PositionFloat types depend on type of the input to the hashgrid.

drjit/hgrid.py
DoeringChristian: Also removed this case, since tcnn was updated.

tests/test_hgrid.py
```diff
@@ -0,0 +1,143 @@
+import drjit as dr
+import drjit.hgrid as hgrid
+import pytest
+import sys
+
+
+@pytest.test_arrays("jit,shape=(*),float16,diff")
+def test01_hash_grid_tcnn(t):
```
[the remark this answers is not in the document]
DoeringChristian: There is now a second test, that uses baked in results from tinycudann. This one no longer relies on that dependency.

tests/test_hashgrid.py
```diff
@@ -0,0 +1,217 @@
+import drjit as dr
+from drjit import hashgrid
```
merlinND, line 2: Remove this line and make sure that we can use `dr.hashgrid` directly (below) without additional imports

drjit/hashgrid.py
```diff
@@ -0,0 +1,721 @@
...
+        This is computed as n_levels * n_features_per_level, representing
+        the concatenation of features from all resolution levels. For example,
+        with 16 levels and 2 features per level, this returns 32.
+        """
+        return self._n_features_per_level * self.n_levels
+
+    def _init_types(self):
+        """
+        Initializes the type aliases used for this hash encoding computation, given a
+        position value ``p``.
```
merlinND, line 275: outdated docstring

drjit/hashgrid.py
```diff
@@ -0,0 +1,721 @@
...
+
+        p = list(p)
+
+        PositionFloat = drjit.leaf_t(p[0])
+
+        PositionFloatXf = (
+            mod.ArrayXf16 if drjit.is_half_v(PositionFloat) else mod.ArrayXf
+        )
+
+        return PositionFloat, PositionFloatXf
```
merlinND, line 304:
```suggestion
        p = list(p)

        PositionFloat = drjit.leaf_t(p[0])
        PositionFloatXf = (
            mod.ArrayXf16 if drjit.is_half_v(PositionFloat) else mod.ArrayXf
        )
        return PositionFloat, PositionFloatXf
```

drjit/hashgrid.py
```diff
@@ -0,0 +1,721 @@
...
+            self._level_offsets[level_i] = offset
+            offset += params_in_level
+
+        self._level_offsets[-1] = offset
+
+        params_size = self._level_offsets[-1] * self.n_features_per_level
+
+        lower = -self._init_scale
+        upper = self._init_scale
+        self.data = drjit.rand(self.dtype, params_size) * (upper - lower) + lower
```
merlinND, line 151: The field `self.data` should probably be created in the constructor. In general, I think it's bad practice so define new fields outside of `__init__()`.

drjit/hashgrid.py
```diff
@@ -0,0 +1,721 @@
...
+        drjit.schedule(self.data)
+
+    def n_params(self) -> int:
+        """
+        The number of parameters, held by this encoding.
+        """
+        return drjit.width(self.data)
```
merlinND, line 158: This should return the correct value even before `alloc()` is called (+ check this in a unit test)

drjit/hashgrid.py
```diff
@@ -0,0 +1,721 @@
...
+            )
+            params_in_level = next_multiple(params_in_level, 8)
+            params_in_level = min(params_in_level, self.hashmap_size)
+
+            self._level_offsets[level_i] = offset
+            offset += params_in_level
+
+        self._level_offsets[-1] = offset
+
+        params_size = self._level_offsets[-1] * self.n_features_per_level
```
merlinND, line 147:
```suggestion
        params_count = self._level_offsets[-1] * self.n_features_per_level
```

merlinND (commented):
Small comments that came up when trying to use the hashgrid.

drjit/hashgrid.py
```diff
@@ -0,0 +1,728 @@
+from __future__ import annotations
+import sys
+import drjit
```
wjakob, line 3: How about `import drjit as dr`? We generally use that prefix, and it makes things look consistent.

drjit/hashgrid.py
```diff
@@ -0,0 +1,728 @@
...
+    This class serves as the interface for Hash based encodings.
+    """
+
+    # Fields, initialized by the ``alloc`` method.
+    data: None | drjit.ArrayBase
+    dtype: None | Type
+    _level_offsets: None | List[int]
+
+    # Fields, defining the hash encoding layout.
+    _dimension: int
```
wjakob, line 38: Needs a 1-line comment per field explaining what it does. (Even if somewhat redundant)

drjit/hashgrid.py
```diff
@@ -0,0 +1,728 @@
...
+    _base_resolution: int
+    _per_level_scale: float
+    _log2_per_level_scale: float
+    _align_corners: bool
+    _torchngp_compat: bool
+    _smooth_weight_gradients: bool
+    _smooth_weight_lambda: float
+    _init_scale: float
+
+    DRJIT_STRUCT = {
```
wjakob, line 51: Is it important for this class to be traversable? Otherwise we can perhaps remove this.

drjit/hashgrid.py
```diff
@@ -0,0 +1,728 @@
...
+        "_per_level_scale": float,
+        "_align_corners": bool,
+        "_torchngp_compat": bool,
+        "_smooth_weight_gradients": bool,
+        "_smooth_weight_lambda": float,
+        "_init_scale": float,
+        "_n_params": int,
+    }
+
+    def __init__(
```
wjakob, line 69: `__init__` Needs a docstring.

drjit/hashgrid.py
```diff
@@ -0,0 +1,728 @@
...
+
+
+def next_multiple(num: int, multiple: int) -> int:
+    """Round `num` to the next multiple of `multiple`"""
+    return multiple * div_round_up(num, multiple)
+
+
+class HashEncoding:
+    """
+    This class serves as the interface for Hash based encodings.
```
wjakob, line 29: Docstring a bit short. You could say that this is the base class of :py:class:`drjit.nn.HashGrid` and  :py:class:`drjit.nn.PermutoEncoding`, and that it stores various fields that are common to both of them.

drjit/hashgrid.py
```diff
@@ -0,0 +1,728 @@
...
+        self._torchngp_compat = torchngp_compat
+        self._smooth_weight_gradients = smooth_weight_gradients
+        self._smooth_weight_lambda = smooth_weight_lambda
+        self._init_scale = init_scale
+
+        # Compute parameter counts per level as well as level offsets, using
+        # dense grids for small scales and hash tables (limited by hashmap_size)
+        # for larger scales.
+
+        # TODO: n_features should be a multiple of 2, and if we add a logging mechanism
```
wjakob, line 101: You can use
```
import warnings
warnings.warn("Issue ...", RuntimeWarning)
```

drjit/hashgrid.py
```diff
@@ -0,0 +1,728 @@
...
+            self.hashmap_size % 8
+        ) == 0, f"Invalid hashmap size {self.hashmap_size}, must be a multiple of 8."
+
+        self._level_offsets = [None] * (self.n_levels + 1)
+        max_params = drjit.scalar.UInt32(0xFFFFFFFF) // 2
+        offset = 0
+
+        for level_i in range(self.n_levels):
+            res = self._resolution(self._level_scale(level_i))
+            stride = drjit.power(float(res), self.dimension)
```
wjakob, line 114: You can use the `**` operator, this isn't really Dr.Jit code. Why `float`?

drjit/hashgrid.py
```diff
@@ -0,0 +1,728 @@
...
+        upper = self._init_scale
+        self.data = (
+            drjit.rand(self.dtype, self.n_params, seed=seed) * (upper - lower) + lower
+        )
+        drjit.schedule(self.data)
+
+    @property
+    def n_params(self) -> int:
+        """
+        The number of parameters, held by this encoding.
```
wjakob, line 167: remove comma

drjit/hashgrid.py
```diff
@@ -0,0 +1,728 @@
...
+
+        return result
+
+    def _alloc(
+        self, dtype: Type[drjit.ArrayBase], seed: int | drjit.AnyArray | None = None
+    ) -> None:
+        """Initialize data storage and compute memory layout for hash grid levels.
+
+        This function:
+        1. Sets up DrJit array types based on the requested dtype
```
wjakob, line 150: numbering seems off?

drjit/hashgrid.py
```diff
@@ -0,0 +1,728 @@
...
+    @property
+    def n_params(self) -> int:
+        """
+        The number of parameters, held by this encoding.
+        """
+        return self._n_params
+
+    def set_params(self, values: drjit.ArrayBase, copy=False) -> None:
+        """
+        Should pass `values` as a DrJit array.
```
wjakob, line 173: For docstring: What purpose does this function serve?

wjakob (commented):
This looks amazing 😍. Here is some minor feedback from me.

drjit/hashgrid.py
DoeringChristian: I will remove all except for "data", as this has to be traversable for frozen functions.

wjakob (commented):
The implementation looks great now, just some minor wording suggestions for a comment.

The main remaining issues I can see are:

- no documentation in `nn.py`
- the docstrings/binding documentation isn't included in the reference

Please add docs, generate them in Sphinx.

wjakob:
@DoeringChristian @merlinND It's merged! 🎉 . Thank you for your hard work on this!!!

I will do some minor text-smithing as a follow-up commit.

PR #398, opened 2025-05-27
pcg32: reverse number generation

lnuic:
Implements efficient reverse-number generation for PCG32.
Reference: [Reversible random number generation](http://robotics.ucsd.edu/ReversibleRNG.pdf)


wjakob:
This looks great!

PR #402, opened 2025-06-12
Fix typo in Adam's update equation

mcrescas:
I have found what I believe to be a discrepancy of Adam's update equation in the documentation.
This PR introduces the necessary corrections.


njroussel (approved):
Yes, indeed. Thank you @mcrescas

PR #407, opened 2025-06-19
Frozen Function Kernel History

DoeringChristian:
Adds test checking that frozen functions generate kernel history entries.


ziyi-zhang:
Hi @DoeringChristian,

Another minor thing I encountered with kernel history & freezing is that it keeps warning 
> traverse(): You passed a value of type <class 'NoneType'> to a frozen function, it could not be converted to a Dr.Jit type. Changing this value in future calls to the frozen function will cause it to be re-traced. The value is located at ...

This repeats in every iteration when the frozen function is inside a loop.

Best

merlinND:
Hi @ziyi-zhang,

Thanks for reporting this! For my understanding, do you actually pass an argument equal to `None` to your frozen function, or the log is incorrect?

Also, we could probably lower the log level for this message to `Debug` @DoeringChristian.

DoeringChristian:
Hi @ziyi-zhang,

Thanks for catching that. It looks like almost all calls to a frozen function result in this message. One reason is that I set the return value of the state function to None if no state function was provided. Another case is, that the globals, captured as part of the closure includes modules such as drjit, which also cannot be traversed.
I will change the log level to Debug for now, and fix the state function case.

PR #408, opened 2025-06-25
Submodule update

dvicini:
Update drjit-core and nanobind. The latter is to make the nanobind version consistent again between mitsuba & drjit


njroussel (approved):
LGTM, thanks :+1:

PR #411, opened 2025-07-02
Fix Adam optimizer for half-precision parameters

njroussel:
In the Adam optimizer the EMA debiasing term was explicitly always set to be of type `Float32`. This is an issue when the parameter being optimized is not `Float32` but `Float16` for example, as the learning rate cache checks that its output type matches the parameter's gradient type.

### DIFF
--- a/drjit/opt.py
+++ b/drjit/opt.py
@@ -965,13 +965,15 @@ class Adam(Optimizer[Tuple[int, dr.ArrayBase, dr.ArrayBase]]):
         # Compute the step size scale, which is a product of
         # - EMA debiasing factor
         # - Adaptive/parameter-specific scaling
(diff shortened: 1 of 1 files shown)


wjakob:
In general, to train fp16, you will want to keep the whole optimizer state in fp32. I don't think it makes much sense to have a value in fp16 and momentum in fp16. So I am fine with this change.

njroussel:
Ok :+1:  I didn't want to mess with the numerical precision so I had originally made it consistent with the original implementation (prior to regression).

I've now pushed a new commit. The momentum, step and next value are all computed in fp32 (the next value is then downcast to fp16). For double precision parameters, everything will be done in double precision.

PR #414, opened 2025-07-18
Add ``dr.mul_wide()`` wide integer multiplication operation

wjakob:
This commit adds the ``dr.mul_wide()`` operation, which performs a wide (e.g. i32 x i32 -> i64) integer multiplication. This is the core operation underlying the Philox pseudorandom number generator.

The commit also exposes ``dr.mul_hi()``, which is similar but only returns the upper part of the result.

They map directly onto the ``mul.wide`` and ``mul.hi`` PTX instructions (or analogous LLVM IR).


njroussel (approved):
LGTM

PR #415, opened 2025-07-18
Per function disabling of frozen function

DoeringChristian:
This PR adds an enabled flag to the frozen function decorator, that can enable/disable kernel freezing per function. This boolean can be changed as well.

### DIFF
--- a/drjit/__init__.py
+++ b/drjit/__init__.py
@@ -2509,6 +2509,7 @@ def freeze(
     warn_after: int = 10,
     backend: Optional[JitBackend] = None,
     auto_opaque: bool = True,
+    enabled: bool = True,
 ) -> Callable[[F], F]:
     ...
 
@@ -2522,6 +2523,7 @@ def freeze(
     warn_after: int = 10,
     backend: Optional[JitBackend] = None,
     auto_opaque: bool = True,
+    enabled: bool = True,
 ) -> F:
     ...
 
@@ -2534,6 +2536,7 @@ def freeze(
     warn_after: int = 10,
     backend: Optional[JitBackend] = None,
     auto_opaque: bool = True,
+    enabled: bool = True,
 ) -> Union[F, Callable[[F2], F2]]:
     """
     Decorator to "freeze" functions, which improves efficiency by removing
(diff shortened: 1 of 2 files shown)


drjit/__init__.py
```diff
@@ -2706,9 +2717,10 @@ def __call__(self, *args, **kwargs):
                 if state_fn is not None:
                     input["state_fn"] = state_fn(*args, **kwargs)
 
-                return self.frozen(
-                    input
-                )
+                if self.enabled:
+                    return self.frozen(input)
+                else:
+                    return self.f(input)
```
merlinND, line 2723:
```suggestion
                    return self.f(*args, **kwargs)
```

drjit/__init__.py
```diff
@@ -2773,7 +2786,10 @@ def __call__(self, *args, **kwargs):
                     }
                 if state_fn is not None:
                     input["state_fn"] = state_fn(self.obj, *args, **kwargs)
-                return self.frozen(input)
+                if self.enabled:
+                    return self.frozen(input)
+                else:
+                    return self.f(input)
```
merlinND, line 2792: Same as above

tests/test_freeze.py
```diff
@@ -3771,3 +3771,33 @@ def func(x):
...
+
+    x = dr.arange(t, 11)
+    assert dr.allclose(func(x), frozen(x))
+
+    assert frozen.n_recordings == 1
+
+    x = dr.arange(t, 12)
+    assert dr.allclose(func(x), frozen(x))
+
+    assert frozen.n_recordings == 1
```
merlinND, line 3803:
```suggestion
    frozen = dr.freeze(func, auto_opaque=auto_opaque, enabled=False)
    assert not frozen.enabled

    x = dr.arange(t, 10)
    assert dr.allclose(func(x), frozen(x))
    assert frozen.n_recordings == 0

    frozen.enabled = True
    x = dr.arange(t, 11)
    assert dr.allclose(func(x), frozen(x))
    assert frozen.n_recordings == 1

    x = dr.arange(t, 12)
    assert dr.allclose(func(x), frozen(x))
    assert frozen.n_recordings == 1
```

Fewer empty lines

merlinND (commented):
Thanks for getting this done so quickly!
I think the way that the function is called when `not enabled` needs to be changed, but otherwise it looks good to me.

drjit/__init__.py
```diff
@@ -2706,9 +2717,10 @@ def __call__(self, *args, **kwargs):
                 if state_fn is not None:
                     input["state_fn"] = state_fn(*args, **kwargs)
 
-                return self.frozen(
-                    input
-                )
+                if self.enabled:
+                    return self.frozen(input)
+                else:
+                    return self.f(*args, **kwargs)
```
merlinND, line 2723: Actually this should be moved to the top of the function so that you can skip `inspect.getclosurevars()` and `state_fn`:

```py
if not self.enabled:
    return self.f(*args, **kwargs)
```


Same below.

PR #416, opened 2025-07-23
Fix `dr.rotate()` to match C++ implementation

merlinND:
The Python implementation `dr.rotate()` did not match the C++ implementation:

https://github.com/mitsuba-renderer/drjit/blob/08ab24364a23d4f3f3f8db948e347baa63a3c6e6/include/drjit/quaternion.h#L343-L347


wjakob:
Yep, the old version indeed looks wrong to me. I suspect nobody used this since the pybind11 -> nanobind port.

Issue #160, opened 2023-06-07
Bug in cbrt function for input values of 0


Microno95:
There is a bug in the cbrt function implementation where if the input value is 0, i.e. the mantissa extracted from the input has a value of exactly 0, the resulting output is exactly `0.337986261`.

This occurs because the polynomial approximation to the cube root of the mantissa gives a value of `0.402389795` which is very far off. This is normally not an issue because for all non-zero values, the range of inputs to the polynomial approximation is in [0.5, 1), but this guarantee is broken for the case of 0.

https://github.com/mitsuba-renderer/drjit/blob/8af56c1d5553c4497dfba0a4409290c4051199b5/include/drjit/math.h#L1446-L1450

Putting that value through the newton-raphson iterations improves the result, but still gives an approximation to the cube root too far away from the true value and thus leads to a final result that is wildly incorrect.

The `python` API obfuscates this issue because it falls back on `math.pow(arg, 1/3)` whenever the individual elements of the input are not an array type. e.g. when used in a mitsuba3 scalar_* variant context.

https://github.com/mitsuba-renderer/drjit/blob/8af56c1d5553c4497dfba0a4409290c4051199b5/drjit/router.py#L2899-L2902

I believe a simple solution would be to change the `select` condition from `isfinite(x)` to `isfinite(x) && neq(x, 0)` here:

https://github.com/mitsuba-renderer/drjit/blob/8af56c1d5553c4497dfba0a4409290c4051199b5/include/drjit/math.h#L1467

njroussel:
Hi @Microno95 

Thank you for the detailed bug report :pray:. I pushed a fix in https://github.com/mitsuba-renderer/drjit/commit/98fb5be8ace4e3c2aace0467f5c7520900ee5b91

PR #423, opened 2025-08-07
Support for excess register spilling to CUDA shared memory

lnuic:
This PR adds support for spilling excess registers to shared memory in the CUDA backend.
For more details, see [#160](https://github.com/mitsuba-renderer/drjit-core/pull/160).

### DIFF
--- a/src/python/docstr.rst
+++ b/src/python/docstr.rst
@@ -6156,6 +6156,17 @@
     When enabled, traversal of complex objects, that usually are opaque to
     loops and conditionals, is enabled.
 
+.. topic:: JitFlag_SpillToSharedMemory
+
+    Enable spilling of excess registers into shared memory.
+
+    This flag activates an optimization that stores registers in shared memory
+    when register pressure is high, reducing the need to spill to slower 
+    local memory. This can improve performance by lowering memory latency on 
+    register-intensive kernels. This flag only applies to the CUDA backend.
+
+    This flag is *enabled* by default.
+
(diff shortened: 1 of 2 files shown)


lnuic:
### Summary

This change introduces a flag that improves performance in CUDA kernels with high register usage by spilling excess registers into shared memory.

### Details

* **Performance impact:** For register-heavy kernels, we observed a **5–10% speedup**, consistent with the findings in [Nvidia blog post](https://developer.nvidia.com/blog/how-to-improve-cuda-kernel-performance-with-shared-memory-register-spilling/).
* **OptiX backend:** The flag does not appear to affect OptiX backend. It is possible that OptiX already enables this optimization internally.
* **Validation:** The feature was tested using a synthetic benchmark with configurable register counts.

### Benchmark

Example benchmark kernel:

```python
@dr.syntax
def work(N, n_registers, n_cycles):
    registers = dr.zeros(dr.cuda.ArrayXf, (n_registers, N))

    value = dr.zeros(dr.cuda.Float, N)
    cnt = dr.zeros(dr.cuda.UInt32, N)
    
    while dr.hint(cnt < n_cycles, mode="symbolic"):
        for i in range(n_registers):
            registers[i] += i
            value += registers[i]
        cnt += 1
    return value
```

Result plot:
register_benchmark_registers


Full benchmark script:  [benchmark.py](https://github.com/user-attachments/files/22347574/benchmark.py)

njroussel (approved):
:100:

PR #425, opened 2025-08-11
Added gradient literal test

DoeringChristian:
This PR expands a test to cover literal gradients for matrix vector multiplications using the new coopvec API, and bumps the drjit-core pointer to the commit fixing this test.


wjakob:
Thanks!

PR #429, opened 2025-08-18
CUDA: bindings for GL interop

merlinND:
Depends on https://github.com/mitsuba-renderer/drjit-core/pull/164.


wjakob:
Thank you Merlin!

PR #432, opened 2025-08-21
Add sphdir python implementation

Speierers:
This patch adds the implementation of `sphdir` for Python (similar to the existing C++ implementation).


wjakob:
This needs to be referenced from the docs (reference.rst)

PR #437, opened 2025-08-21
add ad support and fix bug for c++ repeat/tile operations

lnuic:
- Fixed interface mismatch for repeat on DiffArray
- Implemented ad_var_repeat and ad_var_tile


wjakob:
Hi Lovro -- this looks great, but it still needs a test AFAIK.

PR #451, opened 2025-09-25
Fix C++ transform_compose function

dvicini:
The C++ compose_transform function has a small bug. The translation is set in the last row of the matrix, rather than the last column as it should be. This PR fixes this by directly building the 4x4 matrix with the right entries.

I tested this code in my use case and it works, but there are no unit tests (for the entire file) at the moment.


dvicini:
Great, thanks!

PR #453, opened 2025-09-29
Oklab color space conversion

wjakob:
Optimizing colors in RGB space often produces poor visual results because RGB distances don't match human perception. Perceptual color spaces solve this: their distances align with how humans see color differences. Simply switching to a perceptual norm can visibly improve the quality of many algorithms without other changes.

[Oklab](https://bottosson.github.io/posts/oklab/) [Ottosson, 2020] is a perceptual color space designed specifically for image processing. It improves on CIELAB's hue shift problems and HSV's poor perceptual uniformity, while requiring only two 3x3 matrix multiplications and a cube root.

Key properties of Oklab:
- L, a, b coordinates are perceptually orthogonal
- Smooth, uniform color gradients without hue shifts
- Accurate lightness prediction across all hues
- Differentiable transformations suitable for optimization
- Predictable color mixing behavior

This commit adds two conversion functions:
- ``linear_srgb_to_oklab()``: Convert linear sRGB to Oklab space
- ``oklab_to_linear_srgb()``: Convert Oklab back to linear sRGB

Using these functions for perceptual color optimization is straightforward: convert both colors to Oklab with ``linear_srgb_to_oklab()``, then compute the L2 norm of their difference. This gives a perceptually uniform distance metric. Note that the functions expect *linear* sRGB values. For typical gamma-encoded image data, first apply ``dr.srgb_to_linear()``.

This PR bundles two related changes (one used by the OKLab functions, the other one for sRGB encoding):

1. Added ``drjit.replace_shape_t()`` convenience type trait

   In generic Dr.Jit functions (i.e., ones that are supposed to work with many different inputs), it can be rather tedious to obtain a specific intermediate or output type.

   For example, perhaps a function is called with ``Float`` but then it needs access to an equivalent `Matrix2f` type.

   This commit adds a function that makes this possible--in this case, we can run

   ```
   Matrix2f = dr.replace_shape_t(Float, (2, 2, -1), 'matrix')
   ```

2. Improve ``dr.linear_to_srgb()``, ``dr.srgb_to_linear()``

   This commit adds documentation and testing coverage to the sRGB <-> linear color space conversions.

   The behavior of both functions for the ``clamp=False`` is modified as follows: they are now *odd*, i.e., ``f(-x) = -f(x)``. Displays/compositors with extended color spaces (e.g., Display P3) use this convention to encode "out-of-gamut" colors.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -161,6 +161,8 @@ Miscellaneous operations
 .. autofunction:: copy
 .. autofunction:: linear_to_srgb
 .. autofunction:: srgb_to_linear
+.. autofunction:: linear_srgb_to_oklab
+.. autofunction:: oklab_to_linear_srgb
 .. autofunction:: reorder_threads
 
 Just-in-time compilation
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -261,6 +261,7 @@ _______________________
 .. autofunction:: float32_array_t
 .. autofunction:: float64_array_t
 .. autofunction:: replace_type_t
+.. autofunction:: replace_shape_t
 .. autofunction:: detached_t
 .. autofunction:: expr_t
 .. autofunction:: array_t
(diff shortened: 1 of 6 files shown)


njroussel (approved):
LGTM :rocket:

PR #460, opened 2025-11-24
Advanced Tensor Indexing

DoeringChristian:
This PR makes advanced tensor indexing compatible with PyTorch and NumPy.

When indexing a tensor with an index array, we previously indexed all combinations of the arrays. In the following example, we are indexing the element at `[0, 0]`, `[0, 2]`, `[2, 0]`, `[2, 2]`.
```python
In [1]: import drjit as dr
In [2]: from drjit.auto import TensorXf, Float, UInt32
In [3]: x = TensorXf(dr.arange(Float, 16), (4, 4))
In [4]: x
Out[4]: 
[[0, 1, 2, 3],
 [4, 5, 6, 7],
 [8, 9, 10, 11],
 [12, 13, 14, 15]]
In [5]: x[UInt32(0, 2), UInt32(0, 2)]
Out[5]: 
[[0, 2],
 [8, 10]]
```

In contrast, PyTorch and NumPy handle this case differently. If arrays are used to index into tensors, these frameworks only return the entries at the diagonal coordinates `[0, 0]` and `[2, 2]`. This PR changes the behavior of the indexing function to match that of PyTorch.

```python
In [1]: import drjit as dr
In [2]: from drjit.auto import TensorXf, Float, UInt32
In [3]: x = TensorXf(dr.arange(Float, 16), (4, 4))
In [4]: x
Out[4]: 
[[0, 1, 2, 3],
 [4, 5, 6, 7],
 [8, 9, 10, 11],
 [12, 13, 14, 15]]
In [5]: x[UInt32(0, 2), UInt32(0, 2)]
Out[5]: [0, 10]
```

The PR also adds tests that ensure compatibility with the PyTorch indexing mechanism.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -102,6 +102,13 @@ DrJit 1.4.0 (TBA)
   reverse mode via trajectory replay. See the
   :ref:`documentation <diff_loops>` for details.
 
+- **NumPy-style advanced tensor indexing**: Tensor indexing with multiple
+  integer arrays now follows NumPy/PyTorch semantics. Previously,
+  ``t[arr1, arr2]`` selected all combinations (a grid); it now selects
+  element-wise pairs, matching ``torch`` and ``numpy`` behavior. Non-consecutive
+  array indices (e.g., ``t[arr, :, arr]``) correctly broadcast and move the
+  result dimension to the front of the output shape.
+
 - **NumPy-style array/tensor manipulation and sorting**: This release brings a
   large set NumPy-compatible functions for sorting, reshaping, and reindexing
   arrays and tensors. This includes :py:func:`dr.sort() <sort>`,
--- a/docs/freeze.rst
+++ b/docs/freeze.rst
@@ -630,36 +630,48 @@ possible to reuse the same code as long as trailing dimensions do not change.
    func(t, UInt(1), UInt(1), UInt(1))
 
 Dr.Jit also supports advanced tensor indexing, allowing you to use arrays to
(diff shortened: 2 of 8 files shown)


wjakob:
Hi @DoeringChristian  -- could you rebase this on top of `master`? Thanks!

DoeringChristian:
> Hi @DoeringChristian -- could you rebase this on top of `master`? Thank

wjakob:
Hi Christian,

It's great that the change is now reasonably small. I remember from writing the original version that it is really tricky to get right.

I instructed an agent to try to find slicing/indexing combinations that break this. It found the following counterexample:
```
 import torch
 x = torch.arange(5*6*7*8.).reshape(5,6,7,8)
 i = torch.tensor([0,1,2,3])
 x[:, i, 3, i].shape      # torch: (5, 4)  — drjit gives (4, 5)
 y = torch.arange(5*6*7.).reshape(5,6,7)
 y[:, i, None, i].shape   # torch: (4, 5, 1)  — drjit gives (5, 4, 1)
 ```
(An integer between two index arrays gets wrongly treated as a separator, and a None between them wrongly doesn't.)

The suggested fix is:
```cpp
  } else if (has_advanced && (c.type == Component::Slice ||
                              c.type == Component::None)) {
      in_gap = true;
  }
  ```
  
  Please verify and update as you see fit.

wjakob:
Looks great now. Can you add a changelog entry? After that it's good to merge. (You can do so yourself)

PR #464, opened 2025-12-11
Fixed interface for jit_freeze_discard

DoeringChristian:
Makes use of the new interface for `jit_freeze_discard`.

> The previous implementation of jit_freeze_discard could cause a race condition, since another recording could be started > just before querying the jit_freeze_discarded function. This PR fixes this issue by moving the discarded flag to the 
> recording, and changes the interface of the jit_freeze_discard and jit_freeze_discarded functions.


njroussel (approved):
LGTM

I just rebased the branch and submodule.

PR #476, opened 2026-02-16
Adding FMA requirement for LLVM to documentation in what.rst

hms0411rt:
Follow-up to [#472](https://github.com/mitsuba-renderer/drjit/pull/472)


wjakob:
Thanks!

PR #488, opened 2026-04-16
Add the Muon optimizer

wjakob:
Port the Muon optimizer (_MomentUm Orthogonalized by Newton-schulz_) from Keller Jordan's PyTorch reference implementation. Muon targets the 2D hidden weights of neural networks: it runs SGD with (optionally Nesterov-mixed) momentum and replaces each update with the nearest orthogonal matrix via a quintic Newton-Schulz iteration. AdamW-style decoupled weight decay is supported.


wjakob:
(merged directly)

PR #497, opened 2026-05-07
Extend dr.convolve kernel and boundary handling

bryceyin13:
## Summary

This PR extends `dr.convolve()` in two ways: it adds support for discrete 1D Dr.Jit array kernels, and it makes boundary handling explicit through a new `boundary` parameter.

The existing continuous-filter interface remains available for preset filter names and callable filters. Discrete kernels are selected automatically when the `filter` argument is a 1D Dr.Jit array.

### Discrete Kernel Support

`dr.convolve()` can now take a 1D Dr.Jit array as its filter argument:

```python
kernel = dr.scalar.ArrayXf(1, 2, 1)
y = dr.convolve(x, kernel)
```

The existing `axis` behavior is preserved:

- `axis=None` applies separable convolution along all axes.
- `axis=int` or `axis=tuple[int, ...]` restricts convolution to the selected axes.

The output shape is unchanged and matches the input shape. 

Discrete kernels use the standard convolution orientation, so non-symmetric
kernels are applied in the convolution direction rather than as correlation
kernels.

### Boundary Parameter

This PR adds:

```python
boundary="normalize"
boundary="zero"
```

`boundary="normalize"` is the default. When a kernel is truncated at the boundary, the remaining kernel weights are rescaled so that their sum matches the full kernel sum. In the interior, no normalization is applied: discrete kernel weights are used as provided, and continuous kernel weights are used as sampled. If a unit-normalized kernel is desired, users should normalize the kernel before passing it to `dr.convolve()`.

`boundary="zero"` uses zero padding and does not rescale the truncated kernel weights. In the interior, `boundary="zero"` produces the same result as `boundary="normalize"`. The two modes only differ near boundaries. 

This updated normalization behavior applies to `dr.convolve()`. `dr.resample()` keeps its previous behavior.

### Implementation Details

- Added `boundary: Literal["normalize", "zero"] = "normalize"` to `dr.convolve`.
- Added `@typing.overload` interfaces for the continuous-filter and discrete-kernel forms of `dr.convolve`.
- Added runtime dispatch: preset filter names and callables use the continuous path, while 1D Dr.Jit arrays use the discrete-kernel path.
- Added an internal `Resampler::FilterNormalization` mode to support the new continuous-convolution boundary behavior without changing `dr.resample()`.
- Updated the Python binding for `detail.Resampler` to map `boundary="normalize"` and `boundary="zero"` to the corresponding normalization modes.
- Implemented the discrete-kernel path in Python using Dr.Jit array operations over the flattened source buffer, with standard convolution kernel orientation.
- Kept zero-local-sum handling lazy in the discrete path: it uses Dr.Jit masks and `select()` instead of forcing a Python-side reduction before returning: When `boundary="normalize"` encounters a zero local kernel sum, the continuous and discrete paths handle this slightly differently. The continuous path reuses `detail.Resampler`'s existing behavior and raises an error when the local kernel weight sum is zero. The discrete path avoids a Python-side synchronization check on JIT backends; when the local valid kernel sum is zero, it skips the boundary rescaling at that position, which gives the same value as zero padding there.

### Tests

Updated `tests/test_resample.py` to cover the four main convolution paths:

- Continuous filter with `boundary="normalize"`;
- Continuous filter with `boundary="zero"`;
- Discrete 1D Dr.Jit array kernel with `boundary="normalize"`;
- Discrete 1D Dr.Jit array kernel with `boundary="zero"`.

Additional coverage includes:

- Non-symmetric discrete kernels to verify convolution direction;
- JIT backend coverage for the discrete-kernel path;
- Validation that the discrete-kernel overload only accepts 1D Dr.Jit arrays, including rejection of Dr.Jit tensors and Python lists;
- Discrete convolution over selected tensor source axes.

### DIFF
--- a/drjit/__init__.py
+++ b/drjit/__init__.py
@@ -2764,18 +2764,137 @@ def resample(
     else:
         return value
 
+def _convolve_discrete(source: ArrayT, kernel: ArrayBase,
+                       axis: Union[int, Tuple[int, ...], None],
+                       boundary: Literal["normalize", "zero"]) -> ArrayT:
+    """Implementation detail of drjit.convolve() for discrete 1D kernels."""
+
+    shape = source.shape
+    strides = _compute_strides(shape)
+    ndim = len(shape)
+    source_tp = type(source)
+    value = source.array
+
+    if is_tensor_v(kernel) or not is_array_v(kernel) or depth_v(type(kernel)) != 1:
+        raise TypeError(
+            "drjit.convolve(): a discrete 'filter' must be a 1D Dr.Jit "
+            "array.")
+
+    kernel_shape = kernel.shape
+    if len(kernel_shape) != 1:
+        raise RuntimeError(
+            "drjit.convolve(): a discrete 'filter' must be one-dimensional.")
+
+    kernel_size = kernel_shape[0]
+    if kernel_size == 0:
+        raise RuntimeError(
+            "drjit.convolve(): a discrete 'filter' cannot be empty.")
+
+    if axis is None:
+        axis = tuple(range(ndim))
+    else:
+        axis = _normalize_axis_tuple(axis, ndim, 'axis')
+
+    total = prod(shape)
+    center = kernel_size // 2
+
+    for i in axis:
+        res = shape[i]
+        if res == 0:
+            raise RuntimeError(
+                "drjit.convolve(): source resolution cannot be zero.")
+
+        stride = strides[i]
+        Array = type(value)
+        Accum = float32_array_t(Array) if itemsize_v(Array) <= 4 else Array
+        Index = uint32_array_t(Array)
+        KernelArray = type(kernel)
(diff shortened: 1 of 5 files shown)


wjakob:
Hi @bryceyin13  -- I had a look at this PR. Although the exposed API goes in the direction, the decision to replace the C++ path with an inline Python version does not seem like the right strategy.  Closing in favor of https://github.com/mitsuba-renderer/drjit/commit/00b40a6e873abf2031ed7e11fc19f505b96ec383

bryceyin13:
> Hi @bryceyin13 -- I had a look at this PR. Although the exposed API goes in the direction, the decision to replace the C++ path with an inline Python version does not seem like the right strategy. Closing in favor of [00b40a6](https://github.com/mitsuba-renderer/drjit/commit/00b40a6e873abf2031ed7e11fc19f505b96ec383)

Thanks for taking a look! That makes sense to me.

Issue #498, opened 2026-05-08
Size inference issue for frozen function with literal


dvicini:
There appears to be a size inference issue with frozen functions containing  a literal `0`. The following code produces an array of the wrong size on the second invocation, and hence a mismatch between frozen and regular function call:

```python
import drjit as dr

def func(x, y):
    return x * y, 0.0 * y

x = dr.llvm.Float(1.5)
frozen = dr.freeze(func)

y0 = dr.llvm.Float([10.0, 20.0])
res0 = frozen(x, y0)
ref0 = func(x, y0)
print("i=0: res0[0] size =", dr.width(res0[0]), ", res0[1] size =", dr.width(res0[1]))

y1 = dr.llvm.Float([10.0, 20.0, 30.0])
res1 = frozen(x, y1)
ref1 = func(x, y1)
print("i=1: res1[0] size =", dr.width(res1[0]), ", res1[1] size =", dr.width(res1[1]))

diff = res1[1] - ref1[1] # Size mismatch
```

I am not sure what's the right way to fix this, as I am not familiar with the kernel freezing code.

wjakob:
cc @DoeringChristian

PR #499, opened 2026-05-10
Fix sizes of literals in frozen functions

dvicini:
This addresses #498 

This fix was AI generated, so it needs careful review. The core ideas seem reasonable, but I am not aware of possible pitfalls.

### DIFF
--- a/src/python/freeze.cpp
+++ b/src/python/freeze.cpp
@@ -368,7 +368,7 @@ bool FlatVariables::fill_opaque_mask(FlatVariables &prev,
             (layout_.flags & (uint32_t) LayoutFlag::Literal) &&
             (prev_layout.flags & (uint32_t) LayoutFlag::Literal) &&
             (layout_.literal != prev_layout.literal ||
-             layout_.literal_size != prev_layout.literal_size);
+             this->sizes[layout_.literal_size] != prev.sizes[prev_layout.literal_size]);
 
         opaque_mask[i] |= requires_opaque;
         new_opaques |= requires_opaque;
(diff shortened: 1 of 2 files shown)


DoeringChristian:
Hi @dvicini,

Sorry for the delayed response.
In principle this looks fine. However, we are essentially trading off two different failure conditions. Multiplying the value by 0.0 will simplify it to a literal internally. We therefore loose the information about which input variable the literal size is related to. Adding the literal sizes to the variable the equivalence classes would solve this particular problem, as now the output literal gets tracked as having the same size as the input variable. However, this could introduce other bugs:

```python
import drjit as dr

Float = dr.llvm.Float

def func(x):
    fixed = dr.ones(Float, 3)
    return fixed

frozen = dr.freeze(func)

# Recording: x=3, literal=3 -> same size class
x = Float([1, 2, 3])
res = frozen(x)
ref = func(x)
print(f"record: res_w={dr.width(res)} ref_w={dr.width(ref)}")

# Replay: x=5 -> literal should stay 3, not become 5
x2 = Float([1, 2, 3, 4, 5])
res2 = frozen(x2)
ref2 = func(x2)
print(f"replay: res_w={dr.width(res2)} ref_w={dr.width(ref2)} PASS={dr.width(res2)==dr.width(ref2)}")
```

Because we fundamentally loose the information about input-output relations, for literals, we could decide for one or the other approach. I'm fine with it either way (not sure what causes more confusion)? I tested your implementation for performance. The fix introduces a bunch of hash-set lookups for literals, which make up the largest part of the variables in most scenes. It it seems fine (slight regression on bistro scene, but within noise for my tests):

```
  ┌───────────────┬───────────────────┬───────────────────┐                                                                                                                                                          
  │               │   Fix disabled    │    Fix enabled    │                                                                                                                                                          
  ├───────────────┼───────────────────┼───────────────────┤                                                                                                                                                          
  │ Replay avg    │ 0.0467s ± 0.0029s │ 0.0479s ± 0.0030s │       
  ├───────────────┼───────────────────┼───────────────────┤
  │ Replay min    │ 0.0401s           │ 0.0411s           │                                                                                                                                                          
  ├───────────────┼───────────────────┼───────────────────┤
  │ Speedup (avg) │ 60.04x            │ 58.61x            │                                                                                                                                                          
  ├───────────────┼───────────────────┼───────────────────┤                                                                                                                                                          
  │ No-freeze avg │ 2.8065s ± 0.0232s │ 2.8051s ± 0.0316s │
  ├───────────────┼───────────────────┼───────────────────┤                                                                                                                                                          
  │ Record        │ 2.8504s           │ 2.8713s           │       
  └───────────────┴───────────────────┴───────────────────┘                                                                                                                                                          
```
(measured by profiling 10 times).

Best
Christian

dvicini:
Hi Christian,

I see, the size suddenly becoming "5" in your example also seems very confusing. My fix was purely AI generated, so there is no need to use it as is, if it's not the best solution. Isn't it somehow possible in this test to ensure that the literal stays of the size 3, at which it was explicitly created?

I am not too concerned about the performance hit, as long as we can get correct behavior for various versions of this test.

PR #500, opened 2026-05-18
Butterfly warp-reduction for packet scatter-reduce

lnuic:
Adds `scatter_reduce_packet_dynamic`, the runtime-N analogue of `gather_packet_dynamic`.

### DIFF
--- a/include/drjit/util.h
+++ b/include/drjit/util.h
@@ -353,4 +353,43 @@ void gather_packet_dynamic(size_t packet_size, Source &&source, const Index &ind
     }
 }
 
+/// Special handling of packet scatter-reductions when packet size is only known at runtime
+template <typename Target, typename Value, typename Index, typename Mask>
+void scatter_reduce_packet_dynamic(ReduceOp op, size_t packet_size, Target &target,
+    const Value *values, const Index &index, const Mask &mask_ = true,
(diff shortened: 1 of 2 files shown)


wjakob:
Some (unordered) feedback:

I think that the packet atomics in OptiX aren't supported in all driver versions.
I was surprised that the guards just gate on "ts->compute_capability >= 70" -- are you sure that's correct?

Some of the comments feel a bit AI-generated, e.g. the "amortizing .. scaffolding" and "mirrors" .. are mentioned several times. One comment even mentions pixels in the Dr.Jit-Core PTX generation:

>    // Butterfly warp-reduction over the whole packet collapses up to 32
>    // lanes' contributions targeting the same pixel down to a single set

wjakob:
(Updated the comment). Note that this is actually mostly about the Dr.Jit-Core parts.

lnuic:
The `cc >= 70` / `ptx >= 63` gate is for `match.any.sync` (used for pre-aggregation), not for the vector atomics. 
Vector atomics use the same `supports_vector_reduction` check as before; if they're unavailable the butterfly still runs and the leader just emits N scalar atomics. 
Also cleaned up the comments.

wjakob:
Thanks!

PR #501, opened 2026-05-20
packet ops: add scatter/gather/scatter_reduce masking tests

lnuic:
Adds regression tests for mask handling in packet operations, and bumps drjit-core to include the fix for the scatter_packet_reduce mask bug https://github.com/mitsuba-renderer/drjit-core/pull/191.


wjakob:
That's a lot of new tests just to verify mask handling! Can this not be fused into the existing test with a 1-liner somewhere?

lnuic:
Good point, I have updated the existing tests to include the masking checks.

wjakob:
Perfect, thanks!

PR #502, opened 2026-05-23
SymPy integration

DoeringChristian:
This PR adds integration to sympy using the `dr.wrap` decorator.

[SymPy](https://www.sympy.org/en/index.html) is a symbolic expressions engine. This PR adds it as a target for `dr.wrap`, allowing us to compute higher order derivatives (hessians) etc in Dr.Jit. In contrast to other targets for this decorator the SymPy target does not exchange data between the libraries at runtime, but rather traces the symbolic expressions generated by the wrapped function, and compiles it to Dr.Jit. The wrapped function therefore acts like an ordinary Dr.Jit function, while we can use the complicated symbolic operations supported by SymPy. The wrapped functions are restricted to per-lane operations. Inverting a `Matrix4f` inside of this wrapper will generate the necessary code for Dr.Jit to invert the matrix. Calling the function with a Matrix array in Dr.Jit will therefore invert each matrix per SIMD lane.

### DIFF
--- a/docs/interop.rst
+++ b/docs/interop.rst
@@ -10,7 +10,9 @@ Currently, the following ones are officially supported:
 
 - `NumPy <https://numpy.org>`__,
 - `PyTorch <https://pytorch.org>`__,
-- `JAX <https://jax.readthedocs.io/en/latest/installation.html>`__.
+- `JAX <https://jax.readthedocs.io/en/latest/installation.html>`__,
+- `SymPy <https://www.sympy.org>`__ (symbolic computation, see :ref:`below
+  <interop_sympy>`).
 
 There isn't much to it: given an input array from another framework, simply
 pass it to the constructor of the Dr.Jit array or tensor type you wish to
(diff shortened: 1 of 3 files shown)


docs/interop.rst
```diff
@@ -105,6 +107,130 @@ larger PyTorch program:
...
+  :py:class:`UInt32 <drjit.auto.ad.UInt32>`, …) become ``sp.Symbol(real=True)``.
+  Note that SymPy does not distinguish integer and floating-point symbols—both
+  map to the same kind of symbol. Integer-specific operations like floor
+  division and modulo will produce SymPy expressions (``floor(x/2)``,
+  ``Mod(x, 3)``) that are correctly compiled to Dr.Jit code, but SymPy
+  treats all values as real numbers.
+- Static vectors (:py:class:`Array2f <drjit.auto.ad.Array2f>`,
+  :py:class:`Array3f <drjit.auto.ad.Array3f>`, …) become ``sp.Matrix``
+  column vectors. This means ``.dot()``, ``.cross()``, and ``.norm()`` work
+  immediately. Indexing (``v[0]``), slicing (``v[0:2]``), and unpacking
```
wjakob: "Immediately" seems a bit too excited for a technical documentation.

docs/interop.rst
```diff
@@ -105,6 +107,130 @@ larger PyTorch program:
...
+  (``x, y, z = v``) also work, so existing code that treated vectors as lists
+  continues to work unchanged.
+- Dynamic arrays (:py:class:`ArrayXf <drjit.auto.ad.ArrayXf>`) also become
+  ``sp.Matrix`` column vectors. Their size is known when the function is first
+  called and used for compilation.
+- Matrix types (:py:class:`Matrix3f <drjit.auto.ad.Matrix3f>`,
+  :py:class:`Matrix4f <drjit.auto.ad.Matrix4f>`, …) become square
+  ``sp.Matrix`` objects.
```
wjakob, line 166: I guess tensors aren't supported? Might be worth saying so.

docs/interop.rst
```diff
@@ -105,6 +107,130 @@ larger PyTorch program:
...
+
+Caching
+^^^^^^^
+
+Compiled functions are cached at two levels. An in-memory cache, keyed by the
+argument type signature, avoids recompilation when the same function is called
+repeatedly with arguments of the same types. A disk cache in
+``$HOME/.drjit/sympy/`` persists compiled bytecode across interpreter
+restarts, following Dr.Jit's kernel cache convention (``$HOME/.drjit/``). To
```
wjakob: What about windows? It doesn't use the $HOME convention. Does the cache end up at the right place there too?

drjit/sympy/custom.py
```diff
@@ -0,0 +1,103 @@
+import sympy as sp
+from sympy.utilities.iterables import is_sequence
+
+
+# Custom hessian implementation with ``evaluate = False`` default
+def hessian(f, varlist, constraints=(), evaluate: bool = False):
```
wjakob: This looks like a remainder maybe from your project? The function isn't part of the documented API -- should it be deleted?

drjit/sympy/custom.py
```diff
@@ -0,0 +1,103 @@
...
+            out[i + m, j + m] = f.diff(varlist[i], evaluate=evaluate).diff(
+                varlist[j], evaluate=evaluate
+            )
+    for i in range(N):
+        for j in range(i + 1, N):
+            out[j, i] = out[i, j]
+    return out
+
+
+class StopGradient(sp.Function):
```
wjakob: This looks like a remainder maybe from your project? The function isn't part of the documented API -- should it be deleted?

drjit/sympy/__init__.py
```diff
@@ -0,0 +1,12 @@
+"""drjit.sympy: SymPy-DrJit compatibility layer.
+
+Provides automatic compilation of SymPy expressions to DrJit code,
+custom SymPy functions for DrJit interop, and the ``@dr.wrap(source='drjit',
+target='sympy')`` integration.
+
+Requires: sympy, xxhash, tqdm (optional dependencies of drjit).
```
wjakob: It would be nicer not to depend on tqdm since many people will not have that installed. Can we also avoid depending on xxhash as a python library? We already ship it as part of Dr.Jit on the C side. Is it even important that we use specifically this hash or could it be any Python hash?

drjit/sympy/generator.py
```diff
@@ -0,0 +1,249 @@
...
+
+    def _print_exp(self, expr):
+        """Print exponential function using dr.jit"""
+        return f"dr.exp({self._print(expr.args[0])})"
+
+    def _print_sqrt(self, expr):
+        """Print square root using dr.jit"""
+        return f"dr.sqrt({self._print(expr.args[0])})"
+
+    def _print_sin(self, expr):
```
wjakob: What about overriding `_print_Function` once and then catching the different functions there? I think that's cleaner than providing so many different overrides (and also seems somehow more stable).

drjit/sympy/wrapper.py
```diff
@@ -0,0 +1,479 @@
...
+        print_code: Whether to print generated DrJit code.
+        print_expr: Whether to print the SymPy expression.
+        cache: Dictionary mapping argument signatures to compiled DrJit functions.
+    """
+
+    def __init__(
+        self,
+        f_sp,
+        key_fn: Callable[[...], sp.Expr] | None = None,
+        print_code: bool = False,
```
wjakob: I don't recall reading about these features in the documentation (`print_*`, `save_expr`, `simplify`, `clear_derivatives`). I would prefer having a more minimal implementation without extra features unless they are really useful and needed (and then they should be documented).

drjit/sympy/wrapper.py
```diff
@@ -0,0 +1,479 @@
...
+                print("Expression:")
+                sp.pprint(expr)
+
+            if self.save_expr:
+                # Write out the expression for debugging purposes
+                (cache_dir / f"{key}.expr").write_text(sp.pretty(expr, wrap_line=False))
+
+            expr = FlatOutput(expr)
+
+            # Evaluate expressions and clear unprintable expressions such as
```
wjakob: Unclear where derivatives come from here. Are those derivatives within SymPy?

tests/test_sympy.py
```diff
@@ -0,0 +1,533 @@
...
+
+    # simple
+    assert "dr.select" in p.doprint(sp.Piecewise((x, x > 0), (0, True)))
+
+    # nested
+    expr = sp.Piecewise((x, x > 1), (y, x > 0), (0, True))
+    assert p.doprint(expr).count("dr.select") == 2
+
+
+def test06_printer_dirac_delta():
```
wjakob: The only place where a dirac delta makes sense AFAIK is within an integral where it then collapses the integral. To me it seems wrong to convert it into valid code (including a select statement).

tests/test_sympy.py
```diff
@@ -0,0 +1,533 @@
+"""Tests for drjit.sympy — SymPy-DrJit compatibility layer.
+
+Covers the DrJit code printer, code generation, Hessian/StopGradient utilities,
+@dr.wrap(source='drjit', target='sympy') integration, and automatic type
+promotion.
+
+Requires: sympy, xxhash, tqdm
+"""
+
+import sympy as sp
```
wjakob: Like with other backends for `@dr.wrap`, we should not add a required dependence on sympy, so all of the tests should be skipped of sympy is unavailable.

tests/test_sympy.py
```diff
@@ -0,0 +1,533 @@
+"""Tests for drjit.sympy — SymPy-DrJit compatibility layer.
+
+Covers the DrJit code printer, code generation, Hessian/StopGradient utilities,
+@dr.wrap(source='drjit', target='sympy') integration, and automatic type
+promotion.
+
+Requires: sympy, xxhash, tqdm
```
wjakob: Let's avoid the extra dependencies (xxhash, tqdm).

docs/interop.rst
```diff
@@ -105,6 +107,130 @@ larger PyTorch program:
...
+
+   >>> x = Float(3.0)
+   >>> dr.enable_grad(x)
+   >>> @dr.wrap(source='drjit', target='sympy')
+   ... def f(x):
+   ...     return x**2 + 2*x + 1
+   >>> y = f(x)
+   >>> dr.backward(y)
+   >>> x.grad
+   [8]
```
wjakob, line 146: It would be nice to have a more interesting example here. e.g. to solve an equation, compute an nth-derivative or something like that.

docs/interop.rst
```diff
@@ -105,6 +107,130 @@ larger PyTorch program:
...
+
+Parallelism
+^^^^^^^^^^^
+
+Recall that every Dr.Jit type is inherently an *array* of the concept it
```
wjakob: This section seems a bit long-winded. I guess the parallelize over elements part of Dr.Jit should be quite natural once people reach this part of the documentation.

wjakob:
Hi Christian,

thank you for the PR, this looks like a very useful feature. My suggestion would be to do another crunching pass over the code. 1.5K is .. a lot! I feel like this code has some of the fluffiness of partly AI-generated and could do with some simplification passes to reduce it to a more minimal form (matching also the documented API surface).

Thanks,
Wenzel

wjakob (commented):
I realize I never submitted my comments from the original PR :-(

drjit/sympy/wrapper.py
DoeringChristian: SymPy often leaves derivatives unevaluated for example `floor` is piecewise differentiable, but since it is not at any whole number, SymPy will emit `sp.diff(sp.floor(x)) = sp.Derivative(sp.floor, x)`. In most applications the user will want it to be 0. I removed the automatic handling and documented it this case.

tests/test_sympy.py
DoeringChristian: Similar to the derivatives. SymPy will emit dirac deltas when you try to differentiate step functions. These can only be handled within SymPy itself, and so I set them to 0. The user can also handle this at the end of the function themselves, which is now the case (documented).

PR #506, opened 2026-06-03
Disable JAX GPU preallocation in tests

lnuic:
JAX preallocate most of the GPU by default, starving the cuda backend so the cuda wrap tests fail with out of memory. Disable it via XLA_PYTHON_CLIENT_PREALLOCATE=false
https://docs.jax.dev/en/latest/gpu_memory_allocation.html

### DIFF
--- a/tests/test_wrap.py
+++ b/tests/test_wrap.py
@@ -1,3 +1,4 @@
+import os
 import warnings
 
 import drjit as dr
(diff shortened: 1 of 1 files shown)


lnuic:
I have updated the PR based on the recent drop of TensorFlow support.

wjakob:
Thanks!

PR #509, opened 2026-06-11
Default riscv builds to scalar mode

carlosqwqqwq:
## Why

Dr.Jit supports aggressive x86 and Arm vectorization and optional JIT backends, but it does not currently ship a native RISC-V packet or JIT backend. In simulated or cross-target `riscv64` validation built from an `x86_64` host compiler, the platform detection in `fwd.h` can incorrectly inherit host `x86_64` SIMD assumptions. In addition, CMake still defaults to building LLVM/CUDA/Metal backends that are not available for a conservative RISC-V portability build. This patch makes RISC-V explicit and defaults it to the scalar header-only path.

## What changed

- Add explicit `DRJIT_RISCV_64` / `DRJIT_RISCV_32` detection in `include/drjit/fwd.h`.
- Make RISC-V take precedence over host `x86_64` / Arm predefined macros during simulated or cross-target validation.
- Update `CMakeLists.txt` so `riscv*` targets default `DRJIT_ENABLE_LLVM`, `DRJIT_ENABLE_CUDA`, and `DRJIT_ENABLE_METAL` to `OFF`.
- Emit a status message that RISC-V currently uses the scalar header-only path because no RISC-V JIT backend is available.
- Guard JIT-only install targets so a no-JIT configuration remains valid.

## Verification

- Ran `git submodule update --init --recursive` to fetch `drjit-core`, `nanobind`, and nested dependencies required by the build.
- Ran native CMake configuration with Ninja using `DRJIT_ENABLE_LLVM=OFF`, `DRJIT_ENABLE_CUDA=OFF`, `DRJIT_ENABLE_METAL=OFF`, `DRJIT_ENABLE_PYTHON=OFF`, `DRJIT_ENABLE_AUTODIFF=OFF`, and `DRJIT_ENABLE_TESTS=OFF`.
- Built the native tree successfully with `cmake --build build-codex-native-060 --parallel 4`, which completed with `ninja: no work to do.` for the interface-only configuration.
- Ran simulated `riscv64` CMake configuration with `CMAKE_SYSTEM_NAME=Linux`, `CMAKE_SYSTEM_PROCESSOR=riscv64`, `CMAKE_TRY_COMPILE_TARGET_TYPE=STATIC_LIBRARY`, forced `__riscv=1` / `__riscv_xlen=64`, and `DRJIT_ENABLE_PYTHON=OFF`, `DRJIT_ENABLE_AUTODIFF=OFF`, `DRJIT_ENABLE_TESTS=OFF`.
- Confirmed from `build-codex-riscv-060/CMakeCache.txt` that `DRJIT_ENABLE_LLVM=OFF`, `DRJIT_ENABLE_CUDA=OFF`, and `DRJIT_ENABLE_METAL=OFF` on the simulated `riscv64` target.
- Preprocessed `include/drjit/fwd.h` under forced `__riscv=1` / `__riscv_xlen=64` and confirmed `DRJIT_RISCV_64=1` is defined while host `x86` and Arm SIMD target macros are not.
- Compiled a smoke-test translation unit under forced `__riscv=1` / `__riscv_xlen=64` that includes `<drjit/array.h>` and `static_assert`s the default `drjit::Array<float>` packet size falls back to scalar `Size == 1`.

## Notes

This is a conservative portability patch. It does not implement a RISC-V packet backend, RVV vectorization backend, or RISC-V JIT backend. The goal is to ensure that RISC-V targets do not inherit host `x86_64` assumptions and that Dr.Jit can be configured into an explicit scalar header-only mode for future RISC-V work.


wjakob:
Hi @carlosqwqqwq, sorry, I have no plans to support Risc-V in this project for now until this platform becomes way more mainstream.

PR #514, opened 2026-06-22
Fix minor metal warning

Speierers:
Remove unused `ResultSize` capture from the `to_tuple` lambda in `tex_eval_fetch`. Since `ResultSize` is `constexpr`, it as constant initialization and does not need to be captured by the lambda, it is already usable inside the body via name lookup.

This silences a `-Wunused-lambda-capture` warning emitted by Clang for every instantiation of `bind_texture<Type, Dimension>` (3 dtypes × 3 dimensions × 3 result types = 27 instances per translation unit), considerably reducing build noise in `scalar.cpp` and other Python binding TUs.

### DIFF
--- a/src/python/texture.h
+++ b/src/python/texture.h
@@ -44,7 +44,7 @@ static nb::object tex_eval_fetch(const Tex &texture,
                                  const dr::Array<T, Dimension> &pos,
                                  const dr::mask_t<T> &active) {
     constexpr size_t ResultSize = 1 << Dimension;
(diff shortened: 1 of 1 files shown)


wjakob:
Hi Sébastien -- removing that contexpr here then causes MSVC to fail with a compilation error (that's how the code was written at some intermediate point, which causes CI failures). Can you propose a variant that makes both MSVC and Clang happy?

wjakob:
(NVM, Will fix myself in an upcoming commit. One can just move the `constexpr` by 2 lines).

PR #518, opened 2026-06-27
Fix packet scatter to raw pointers

jeongseok-meta:
## Motivation

`dr::scatter()` accepts raw-pointer targets, but packet values currently try to call `scatter_packet_` on the pointer type. Pointer expressions such as `.data()` also cannot bind to the existing lvalue target parameter. This prevents [Momentum's rasterizer](https://github.com/facebookresearch/momentum/blob/v0.1.114/momentum/rasterizer/image.cpp#L166) from building.

conda-forge currently carries [a patch for this behavior](https://github.com/conda-forge/drjit-cpp-feedstock/blob/9a731565e61ff8d4373c79715ea59a606a05e2d2/recipe/patches/drjit/0004-Fix-scatter-for-raw-pointer-target.patch). It can be removed after this fix is included in a Dr.Jit release.

## Changes

- Preserve the existing lvalue overload and add a constrained overload for pointer expressions.
- Dispatch raw-pointer packet scatters through the outer static array, selecting the generic element-wise fallback; JIT/AD specializations require an array target.
- Normalize scalar and SIMD indices to unsigned 32-bit values without changing their nesting depth, so each recursive packet layer contributes its flattening stride.
- Require a static native outer packet with one clear compile-time diagnostic, without instantiating pointer-incompatible overloads, while preserving native nested `DynamicArray` recursion.
- Cover padded, nested, and deeply nested packet layouts, native nested `DynamicArray` values, signed indices, and explicit-template lvalue calls.

## Testing

- Fresh GCC 15/CMake build: the extension compiled and the four focused raw-pointer packet tests passed, covering scalar, nested, deeply nested, and native nested `DynamicArray` values.
- Standalone scalar, masked, nested, and signed-index packet scatter probes passed.
- A depth-three packet value with a depth-one SIMD index produces `[1, ..., 16]`, confirming recursive flattening without collisions.
- Native nested `DynamicArray` packet scatter produces `[1, 2, 3, 4, 5, 6]`; dynamically sized outer values and JIT/AD packet values fail with only the intended unsupported-value diagnostic.
- The previous vendored broadcast fallback also supports native nested `DynamicArray` values but fails to compile for static arrays containing JIT/AD elements; Momentum's raw-pointer callers use native fixed-size packets.
- `tests/test_call_ext.py`: 128 passed
- Unpatched Dr.Jit reproduces Momentum rasterizer failures at both pointer-expression and packet-dispatch call sites.
- With this change, Momentum's rasterizer and pymomentum renderer build, and `pymomentum/test/test_renderer.py` passes all 8 tests.
- `python -m py_compile tests/test_memop_ext.py`
- `git diff --check`

### DIFF
--- a/include/drjit/array_router.h
+++ b/include/drjit/array_router.h
@@ -1109,10 +1109,9 @@ void scatter(Target &target, const Value &value, const Index &index,
                 // Use the outer static array to select the generic element-wise
                 // fallback; JIT/AD packet scatter requires an array target.
                 using ValueD = std::decay_t<Value>;
-                constexpr bool Supported =
-                    ValueD::Size != Dynamic && !is_jit_v<ValueD>;
+                constexpr bool Supported = !is_dynamic_v<ValueD>;
                 static_assert(Supported,
-                              "Raw-pointer packet scatter requires a statically sized non-JIT value");
+                              "Raw-pointer packet scatter requires a fully statically sized value");
                 if constexpr (Supported) {
                     ValueD::template scatter_packet_<ValueD::Size>(
                         target, value, uint32_array_t<value_t<Value>>(index),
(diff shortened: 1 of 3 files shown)


jeongseok-meta:
The four red TeamCity contexts did not reach a build: each reports failedToStart because the untrusted fork-build approval timed out. Could a maintainer approve or retrigger them when convenient? The head remains mergeable, and the focused local validation in the PR description is still current.

PR #521, opened 2026-06-27
Fix dr.dot on cooperative vectors using the wrong operand

lnuic:
Fix dr.dot on cooperative vectors silently computing dot(b, b) instead of dot(a, b) (first operand was dropped), and add a regression test.


wjakob:
Thanks!

PR #522, opened 2026-06-27
Fix cumulative sum over a tuple of axes only applying one axis

lnuic:
The tuple-axis loop re-reduced the original input instead of accumulating, so only the last axis was applied. Added a regression test.


wjakob:
Thanks!

PR #523, opened 2026-07-10
Fix 2D dynamic slice

dvicini:
There appears to be an edge case 2D slices of dynamic size. This PR adds a fix and a test that triggers the issue on the current master. The test suite previously did not cover slices that are larger than 4 elements and trigger the "dynamic" size case.

I am a bit unsure about the fix (AI + manual inspection), but these days it feels a bit silly to open an issue without trying to attempt to fix.


wjakob:
Thanks Delio, the fix looks good to me.

PR #528, opened 2026-07-28
Fix product reduction gradient when an element is zero

lnuic:
The gradient of a product should be the product of the other entries, but we
computed it as `result / value`, which breaks when an entry is zero. 
`dr.prod` caught the divide by zero and returned 0, which is the wrong answer: for
`[2, 0, 4]` the gradient should be `[0, 8, 0]` and previous result was `[0, 0, 0]`. 

The axis/block version didn't check at all and returned NaN.

The cost is two extra reductions in the backward pass, so a product backward is about 3x slower:

| backend | op | N | before | after |
|---------|----------------|-----|----------|----------|
| LLVM    | `prod`         | 16M | 3.8 ms   | 13.0 ms  |
| LLVM    | `block_reduce` | 16M | 6.0 ms   | 13.0 ms  |
| CUDA    | `prod`         | 16M | 0.044 ms | 0.135 ms |
| CUDA    | `block_reduce` | 16M | 0.090 ms | 0.314 ms |

That could be brought back to a single pass by combining the two reductions, but
it needs a new reduction primitive in drjit-core, which felt like too much for
this one fix. I am happy to add that if needed.

If the slowdown isn't worth it I am happy to leave the gradient approximate and
just guard the block version so it returns 0 instead of NaN.

### DIFF
--- a/src/extra/autodiff.cpp
+++ b/src/extra/autodiff.cpp
@@ -2198,6 +2198,22 @@ struct BlockReduceEdge : Special {
         }
     }
 
+    JitVar mul_weight(size_t size) const {
+        Index i0 = m_value_in.index();
+        JitVar z = scalar(i0, 0.0),
+               o = scalar(i0, 1.0),
+               n_zero = dr::block_sum(dr::select(m_value_in == z, o, z),
+                                      m_block_size, m_symbolic),
+               nz_prod = dr::block_reduce(ReduceOp::Mul,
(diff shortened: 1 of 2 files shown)


wjakob:
Hi Lovro -- are you sure about needing new backend operations? Dr.Jit supports `block_prefix_reduce` which seems to be just what is needed.  What I don't like about this version is that it seems to make the primal slower even when no derivatives are ultimately needed.

PR #529, opened 2026-07-29
Fix reversed operands in cooperative vector __rsub__

lnuic:
`__rsub__` used the same function as `__sub__` without swapping the operands, so `5.0 - vec` quietly evaluated to `vec - 5.0` instead.


wjakob:
ugh.. :). Good catch!

PR #537, opened 2026-08-29
Enable jit variable bounds in dr.binary_search

matttsss:
This commit adds a missing functionality that is present in CPP where the binary search supports jit variable bounds.

It also adds missing typing to the function.

New tests were added as none were present for the `dr.binary_search`.

### DIFF
--- a/drjit/__init__.py
+++ b/drjit/__init__.py
@@ -3733,20 +3733,20 @@ def rng(seed: Union[ArrayBase, int] = 0, method='philox4x32', symbolic: bool = F
         raise RuntimeError("Only generator='philox4x32' is currently supported.")
 
 
-def binary_search(start, end, pred):
+def binary_search(start: Union[ArrayBase, int], end: Union[ArrayBase, int], pred: Callable) -> Union[ArrayBase, int]:
     '''
     Perform a binary search over a range given a predicate ``pred``, which
     monotonically decreases over this range (i.e. max one ``True`` -> ``False``
     transition).
 
-    Given a (scalar) ``start`` and ``end`` index of a range, this function
-    evaluates a predicate ``floor(log2(end-start) + 1)`` times with index
-    values on the interval [start, end] (inclusive) to find the first index
-    that no longer satisfies it. Note that the template parameter ``Index`` is
-    automatically inferred from the supplied predicate. Specifically, the
-    predicate takes an index array as input argument. When ``pred`` is ``False``
-    for all entries, the function returns ``start``, and when it is ``True`` for
-    all cases, it returns ``end``.
+    Given ``start`` and ``end`` indices of a range, this function evaluates a
+    predicate ``floor(log2(end-start) + 1)`` times with index values on the
+    interval [start, end] (inclusive) to find the first index that no longer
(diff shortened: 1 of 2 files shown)


drjit/__init__.py
```diff
@@ -3765,24 +3765,47 @@ def binary_search(start, end, pred):
...
     Returns:
         Index array resulting from the binary search
     '''
-    assert isinstance(start, int) and isinstance(end, int)
+    if depth_v(start) > 1 or depth_v(end) > 1 :
+        raise ValueError(f"`start` and `end` arguments should have depth <= 1, got {depth_v(start)} and {depth_v(end)}")
 
-    iterations = log2i(end - start) + 1 if start < end else 0
+    # Cast type and possibly convert to JIT the second variable
+    tp = type(start + end)
```
wjakob: dr.expr_t does this

drjit/__init__.py
```diff
@@ -3765,24 +3765,47 @@ def binary_search(start, end, pred):
...
 
     Returns:
         Index array resulting from the binary search
     '''
-    assert isinstance(start, int) and isinstance(end, int)
+    if depth_v(start) > 1 or depth_v(end) > 1 :
+        raise ValueError(f"`start` and `end` arguments should have depth <= 1, got {depth_v(start)} and {depth_v(end)}")
 
-    iterations = log2i(end - start) + 1 if start < end else 0
+    # Cast type and possibly convert to JIT the second variable
```
wjakob: > # Cast type and possibly convert to JIT the second variable
Ungrammatical.

I would say: "Transform inputs to a consistent type" or similar

drjit/__init__.py
```diff
@@ -3765,24 +3765,47 @@ def binary_search(start, end, pred):
...
+    tp = type(start + end)
+    start = tp(start)
+    end = tp(end)
 
-    for _ in range(iterations):
+    iterations = detach(select(start < end, log2i(abs(end - start)) + 1, 0))
```
wjakob: I don't understand the `detach`. These are not differentiable variables, and it should not matter whether we access them through `.ad` types or not.

tests/test_binary_search.py
```diff
@@ -0,0 +1,258 @@
+import drjit as dr
+import numpy as np
+import pytest
+
+
+def _ref(start, end, values, thresholds):
```
wjakob, line 6: This kind of test coverage is a bit insane and some of the tests look highly redundant. Is it by any chance AI-generated?

Please use parameterized tests to cover more variants with less code. We don't need many individual tests for obscure corner cases. Often some of them can be rolled into 1 extra line of a combined test.

drjit/__init__.py
matttsss: I wasn't aware, thanks!

drjit/__init__.py
matttsss: There is no reason yes, I should have removed it in the cleanup

tests/test_binary_search.py
matttsss: Yes they are, I already did a cleanup pass on them but I agree they were still a bit long

